PR #3, opened 2018-12-19
Couple of patches for alerts

jamesjefferies:
Comparing a symbol with a string can cause confusion!
Date time formatter not quite right


PhilipTB:
Agreed to changes, look ok, thanks

PR #41, opened 2019-11-21
Meter attributes spike

jamesalmond:


app/services/aggregate_data_service.rb
```diff
@@ -162,11 +164,12 @@ def aggregate_heat_and_electricity_meters
       aggregate_storage_heater_amr_data,
       :storage_heater,
       aggregate_storage_heater_mpan,
-      "#{meter_collection.aggregated_electricity_meters.name} storage heater only"
+      "#{meter_collection.aggregated_electricity_meters.name} storage heater only",
+      :storage_heater_aggreated
```
jamesjefferies: Different names?

app/models/meter_collection.rb
```diff
@@ -30,7 +30,7 @@ class MeterCollection
                 :solar_pv,
                 :grid_carbon_intensity
 
-  def initialize(school, holidays:, temperatures:, solar_irradiation:, solar_pv:, grid_carbon_intensity:)
+  def initialize(school, holidays:, temperatures:, solar_irradiation:, solar_pv:, grid_carbon_intensity:, modifiers: {})
```
jamesalmond: A meter collection now accepts some "modifiers"/meter attributes. These are for meters that we don't know about in the application that get created at a later data (i.e. pseudo meters and aggregate meters).

We can continue to call them meter attributes if we prefer, it just seemed a better name when we were discussing it as we were talking about one not directly related to meters we knew about.

app/models/meter_collection.rb
```diff
@@ -275,6 +276,10 @@ def close_time
     @cached_close_time
   end
 
+  def modifiers(type)
```
jamesalmond: Simple accessor to get modifiers with a certain key, defaulting to an empty hash. The keys of the hash will be the purpose of the new meter that is being created (see examples further down the changes)

app/services/aggregate_data_service.rb
```diff
@@ -116,7 +116,8 @@ def aggregate_heat_and_electricity_meters
       electricity_meter.amr_data,
       :electricity,
       electricity_meter.id,
-      electricity_meter.name
+      electricity_meter.name,
+      :storage_heater_disaggreated_electricity
```
jamesalmond: This new parameter is the name of the attributes we might want to pull in for the new type of meter that is being created.

For example, the attributes in the  MeterCollection might look like this:

```
:storage_heater_disaggregated_electricity => {
      meter_corrections: [
        no_heating_in_summer_set_missing_to_zero: {
          start_toy: TimeOfYear.new(5,25),
          end_toy:   TimeOfYear.new(10, 15)
        }
      ]
    }
```

jamesalmond: (probably not a good example of what they might actually be!)

app/services/aggregate_data_service.rb
```diff
@@ -390,7 +395,7 @@ def aggregate_heat_and_electricity_meters
       number_of_pupils: meter.number_of_pupils,
       solar_pv_installation: meter.solar_pv_setup,
       storage_heater_config: meter.storage_heater_setup,
-      meter_attributes: meter.meter_attributes
+      meter_attributes: meter.meter_attributes.merge(@meter_collection.modifiers(modifier_name))
```
jamesalmond: We're pulling the attributes for the copied meter from the meter collection here using the passed in key.
They're currently merged with the parent meter's attributes as the previous behaviour here was to use the parent meter attributes. There are a few options here depending on how we want to scope attributes for these scenarios:

A) merges _all_ the attributes from the parent meter with the attributes for _all_ solar_pv_consumed_sub_meters
```
meter_attributes: meter.meter_attributes.merge(@meter_collection.modifiers(:solar_pv_consumed_sub_meter))
```
B) just uses the attributes for _all_ solar_pv_consumed_sub_meters
```
meter_attributes: @meter_collection.modifiers(:solar_pv_consumed_sub_meter)
```
C) pulls solar_pv_consumed_sub_meter attributes from the parent meter (allows us to specify attributes that are specific to a particular "real" meter's pseudo meters that might not apply to other meters' sub-meters for that school
```
meter_attributes: @meter_collection.modifiers(:solar_pv_consumed_sub_meter).merge(meter.meter_attributes[:solar_pv_consumed_sub_meter])
```
or D) various combinations of the above

app/models/meter_collection.rb
jamesalmond: We will only pass in the attributes relevant to that school, which allows us to use generic names for the attributes. Each school will have different sets of generic attributes passed in (if relevant).

app/services/meter_collection_factory.rb
```diff
@@ -0,0 +1,82 @@
+class MeterCollectionFactory
```
jamesalmond: This class lets us use a simple data object to build an analytics meter collection. Benefits are:

- Smaller YAML files for download
- Creation of the objects is in the analytics, removing the issue earlier this week where the structure of the class had changed from what the YAML was using.
- clearer interface between the two systems
- we can achieve this without having to significantly change how the MeterCollections or Meters are created, the majority of the analytics code remains unchanged.

app/services/meter_collection_factory.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  def self.build(data)
```
jamesalmond: This method accepts the contents of the YAML file directly for ease of use.

app/services/meter_collection_factory.rb
```diff
@@ -0,0 +1,82 @@
...
+      amr_data.add(reading[:reading_date], OneDayAMRReading.new(meter_data[:external_meter_id], reading[:reading_date], reading[:type], reading[:substitute_date], reading[:upload_datetime], reading[:kwh_data_x48]))
+    end
+    Dashboard::Meter.new(
+      meter_collection:   meter_collection,
+      amr_data:           amr_data,
+      type:               meter_data[:type],
+      identifier:         meter_data[:identifier],
+      name:               meter_data[:name],
+      external_meter_id:  meter_data[:external_meter_id],
+      meter_attributes:   meter_data[:modifiers]
```
jamesalmond: Meters that we know about still hold onto their own attributes

app/services/meter_collection_factory.rb
```diff
@@ -0,0 +1,82 @@
...
+      school_data[:postcode]
+    )
+
+    meter_collection = MeterCollection.new(school,
+                                           temperatures: @temperatures,
+                                           solar_pv: @solar_pv,
+                                           solar_irradiation: @solar_irradiation,
+                                           grid_carbon_intensity: @uk_grid_carbon_intensity,
+                                           holidays: @holidays,
+                                           modifiers: modifiers
```
jamesalmond: These are the global attributes for sub/psuedo/aggregated meters created by the analytics.

test_support/school_factory.rb
```diff
@@ -9,10 +9,10 @@ def initialize
   end
 
   # e.g. meter_collection = load_school(:urn, 123456, :analytics_db) source: or :bathcsv, :bathhacked etc.
-  def load_or_use_cached_meter_collection(identifier_type, identifier, source)
-    return load_aggregated_meter_collection if source == :aggregated_meter_collection
-    return load_validated_meter_collection  if source == :load_validated_meter_collection
-    return load_unvalidated_meter_collection  if source == :load_unvalidated_meter_collection
+  def load_or_use_cached_meter_collection(identifier_type, identifier, source, meter_attributes_overrides: {})
```
jamesalmond: This is a hypothetical implementation of how in the test scripts we might be loading a school from the data downloaded in the app but we want to test some change to the meter attributes. We can pass in the overrides where that would get merged into the downloaded ones.

test_support/school_factory.rb
```diff
@@ -37,16 +37,18 @@ def load_or_use_cached_meter_collection(identifier_type, identifier, source)
...
+    validated_meter_collection = build_meter_collection(validated_meter_data, meter_attributes_overrides: meter_attributes_overrides)
     AggregateDataService.new(validated_meter_collection).aggregate_heat_and_electricity_meters
     validated_meter_collection
   end
 
-  def load_unvalidated_meter_collection
+  def load_unvalidated_meter_collection(meter_attributes_overrides: {})
     school_filename = 'freshford-church-school'
-    unvalidated_meter_collection = load_meter_collections(school_filename, 'unvalidated-meter-collection-')
+    unvalidated_meter_data = load_meter_collections(school_filename, 'unvalidated-data-')
```
jamesalmond: Note: changed file names (files are in a new format now)

test_support/school_factory.rb
```diff
@@ -92,6 +94,21 @@ def load_meter_collections(school_filename, file_type)
...
+    MeterCollectionFactory.new(
+      temperatures: data[:schedule_data][:temperatures],
+      solar_pv: data[:schedule_data][:solar_pv],
+      solar_irradiation: data[:schedule_data][:solar_irradiation],
+      grid_carbon_intensity: data[:schedule_data][:grid_carbon_intensity],
+      holidays: data[:schedule_data][:holidays]
+    ).build(
+      school_data: data[:school_data],
+      amr_data: data[:amr_data],
+      meter_attributes: meter_attributes.merge(meter_attributes_overrides)
```
jamesalmond: Could use `MeterCollectionFactory.build` but wanted it to be clear how the new override attributes were being passed in.

app/services/aggregate_data_service.rb
```diff
@@ -162,12 +164,10 @@ def aggregate_heat_and_electricity_meters
       aggregate_storage_heater_amr_data,
       :storage_heater,
       aggregate_storage_heater_mpan,
-      "#{meter_collection.aggregated_electricity_meters.name} storage heater only"
+      "#{meter_collection.aggregated_electricity_meters.name} storage heater only",
+      :storage_heater_aggreated
     )
 
-    # not ideal but aggregate storage heater meter is virtual and created here so needs to acquire attributes
-    aggregate_storage_heater_meter.set_meter_attributes(MeterAttributes.for(aggregate_storage_heater_mpan, '', :electricity))
```
jamesalmond: Don't need to do this any more as they should be available via  `{:storage_heater_aggreated => {}}` on the meter collection

PR #46, opened 2020-01-16
Extra tariff fields

jamesalmond:


app/models/meter_attributes.rb
```diff
@@ -239,6 +239,55 @@ class EconomicTariff < MeterAttributeTypes::AttributeBase
...
+      ),
+      renewable_energy_obligation: MeterAttributeTypes::Hash.define(
+        structure: {
+          per:  MeterAttributeTypes::Symbol.define(allowed_values: [:kwh]),
+          rate: MeterAttributeTypes::Float.define
+        }
+      ),
+      agreed_capacity: MeterAttributeTypes::Hash.define(
+        structure: {
+          per:  MeterAttributeTypes::Symbol.define(allowed_values: [:kwh]),
```
jamesalmond: @PhilipTB Comments in code suggest this is monthly, put you suggested in Trello that they're all per-kwh?

app/models/meter_attributes.rb
```diff
@@ -239,6 +239,55 @@ class EconomicTariff < MeterAttributeTypes::AttributeBase
     )
   end
 
+  def self.default_tariff_rates
```
jamesalmond: Method to share the extra config between standard and differential tariff meter attribute types

PR #63, opened 2021-01-29
WIP Meteostat first functional test

jhigman:


ldodds:
:+1:

ldodds:
This looks good to me. 

Its not actually testing the rate limiting on the client side though? But I guess we're assuming the Limiter gem is doing the right thing here. Can't think how to test that reliably atm.

jhigman:
Yes, that was a bit tricky to test - I've been testing it by setting the
limits to different values and calling it on the command line - requesting
a big date range, where it makes multiple requests, is slowed down quite a
lot with a limit of e.g. 1 req per second.. (so you can see it working, but
not sure about how to automate that test.)

I've tried another version (in that branch) just using the Faraday
middleware for the 429 retry - it's a cleaner implementation, but again
hard to test (so I've taken out those tests). What do you think? Better to
have more 3rd party gems and a more clunky implementation, but be able to
test it, or rely on Faraday to do the right thing and have cleaner gems and
code?

ldodds:
I think even though we have a bit less control over the testing, I'd opt for minimising number of dependencies and just rely on Faraday.

jhigman:
OK great, that would be my preference too. (Always hard to throw stuff away
when you've spent a while figuring it out though..!)

BTW this branch probably also covers the
https://trello.com/c/yV0gerd3/1199-revise-code-for-managing-errors-and-rate-limiting-for-meteostat-api
ticket in Trello.

I'm thinking of adding a couple of tests on the main MeteoStat class to
make sure we handle errors etc, but then leaving it at that for the moment,
so we can check that the process of running tests in analytics is not going
to cause problems with Philip developing on it..

Also I think he wants the latest version of the analytics to be put on the
test site - shall I do that?

Cheers, Julian

ldodds:
Yes, should be fine to do that.

jhigman:
@ldodds 

I've added a timed test, to check the rate limit behaviour (a bit crude, but confirms that limiting is being applied).

I think this will probably do for this branch (unless you can see some things that need changing?)

ldodds:
:+1:

PR #65, opened 2021-02-02
Added readings method to retrieve power only, in familiar format

jhigman:


jhigman:
- [ ] report true/false from consent calls
- [ ] rename 'file_link'
- [ ] url_date - check time ranges around midnight
- [ ] read inventory - use Faraday to retry
- [ ] read inventory - check returned doc, parse into something useful
- [ ] rename "unit adjust" method
- [ ] catch missing start/end dates
- [ ] 'make one day' method - remove start/end params
- [ ] comment in tariff prices about structure
- [ ] status - check for other error codes
- [ ] status - update returned symbols (unknown, consent_reqd, available)

lib/dashboard/data_sources/n3rgy/n3rgy_data_api.rb
```diff
@@ -11,10 +11,10 @@ class NotAllowed < StandardError; end
     DATA_TYPE_TARIFF = 'tariff'
     DATA_TYPE_PRODUCTION = 'production'
 
-    def initialize(api_key, base_url, debugging)
+    def initialize(api_key, base_url)
+      raise ApiFailure.new("Apikey and Base URL required for N3rgyDataApi") unless api_key.present? && base_url.present?
```
ldodds, line 15: Could add names of the ENV variables that should be set here.

PR #70, opened 2021-02-22
Add rollbar context when exceptions are raised during validation

ldodds:
Most of these changes are to add Factory Bot, and factories to the code to allow some basic unit tests.

The functional change is to catch exceptions when validating meters and, if Rollbar has added the `rollbar_context` accessor, to store the meter mpan before letting the exception continue.

The factories should let us start to add further unit tests to the framework.


spec/factories/one_data_amr_reading_factory.rb
```diff
@@ -0,0 +1,15 @@
+FactoryBot.define do
+  factory :one_data_amr_reading, class: "OneDayAMRReading" do
```
jhigman, line 2: Should this be ""one_day_amr_reading" to match class name? (rather than :one_data..)

PR #71, opened 2021-02-23
Add elements call. Add unit tests

ldodds:


lib/dashboard/data_sources/n3rgy/n3rgy_data_api.rb
```diff
@@ -87,11 +94,15 @@ def error_message(response)
       data = JSON.parse(response.body)
       if data['errors']
         error = data['errors'][0]
-        "#{error['code']} : #{error['message']}"
+        error['message']
+      elsif data['message'] != nil
```
jhigman, line 98: You don't need the " != nil", Ruby style would probably just be "elsif data['message]"..?

PR #75, opened 2021-03-01
Add dcc meter flag to dashboard meter

jhigman:


ldodds:
Could update the `spec/factories/meter_factory.rb` file as well so the attribute can be used within tests?

And maybe a simple test for `Dashboard::Meter` even if its just a stub so we can slowly improve coverage?

ldodds:
:+1:

PR #97, opened 2021-06-24
change default electric price from 12p to 15p

ldodds:


ldodds:
Merged into `aws-eb-test` for deployment

PR #102, opened 2021-06-30
Excess availability charge calculation

PhilipTB:
Based on previous branch excess-availability-charge which just defined the meter tariff attribute, this branch does and presents the calculation as well


ldodds:
Merged into aws-eb-test branch for review

PR #104, opened 2021-07-07
Add download_by_component for app

ldodds:
Add a method for the app to call to fetch the rtone data by component, returning a consistent data structure.


ldodds:
Merged into aws-eb-test branch for review

PR #106, opened 2021-07-08
have tuned on error checking for tariffs, but relaxed for older accounting tariffs which I want to deprecate

PhilipTB:


ldodds:
Merged into `aws-eb-test` branch for review

PR #108, opened 2021-07-09
Raise exception for API error

ldodds:
Raise error when there's an error code from RBee API


ldodds:
Merged into aws-eb-test for review

PR #110, opened 2021-07-09
Billing updates 9 jul2021

PhilipTB:
- differential tariffs that span midnight
- tooltips on billing table headers
- made the coverage of tariffs clearer, so refer user back to mail@ if not full coverage in most recent 25 months


ldodds:
Merged and released to on Friday.

PR #112, opened 2021-07-12
Fetch start_date so we can use it to create valid range for loop

ldodds:


lib/dashboard/data_sources/low_carbon_hub_meter_readings.rb
```diff
@@ -30,9 +30,10 @@ def download(meter_id, urn, start_date, end_date)
 
   def download_by_component(meter_id, component, synthetic_mpan, start_date, end_date)
     raw_data = @rbee.smart_meter_data_by_component(meter_id, start_date, end_date, component)
+    actual_start = start_date.nil? ? first_meter_reading_date(meter_id) : start_date
```
jhigman, line 33: Just a Ruby style thing, this could probably be shortened to 

    actual_start = start_date || first_meter_reading_date(meter_id)

jhigman (approved):

PR #118, opened 2021-07-14
Billing tooltips v3

PhilipTB:
I have made the table scrollable, to see if that helps.

Would prefer not to re-orientate table as  you can't scan similar size values and totals at bottom aren't helpful, lots of other reasons which can be discussed.

We we see what this looks like on test. Looks ok in static HTML. Might also want to review, without scrollability, on a standard EDF complex tariff as there won't be quite so many columns.

Derived from last billing tooltips branch, but with table formatting from the last baseload branch which I don't think has been integrated in yet.

### DIFF
--- a/app/models/tariffs/meter_tariff_description.rb
+++ b/app/models/tariffs/meter_tariff_description.rb
@@ -5,6 +5,8 @@ class MeterTariffDescription
       rate(attribute_type)
     when /^Duos.*$/
       duos(attribute_type, school, meter)
+    when /^Climate.*$/
+      'Climate change levy - reflecting your carbon emissions'
     else
       nil
     end
--- a/app/models/tariffs/meter_tariff_description.rb
+++ b/app/models/tariffs/meter_tariff_description.rb
@@ -1,4 +1,7 @@
 class MeterTariffDescription
+  include Logging
+  class UnknownDuosBand < StandardError; end
+
   def self.description_html(school, meter, attribute_type)
     case attribute_type
     when /^\d\d:\d\d to \d\d:\d\d$/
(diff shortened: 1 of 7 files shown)


ldodds:
@PhilipTB I've merged this into the aws-eb-test branch. I had to deal with some merge conflicts because I'd already incorporated and released the baseload branch. Generally its easier if you just push bug fixes for a PR to the same branch, rather than creating another branch.

Can you confirm which of these 3 will need to be merged with master? Or can we close the v2 tooltips? Hard for me to know without reviewing all the changes across all branches.

PhilipTB:
> @PhilipTB I've merged this into the aws-eb-test branch. I had to deal with some merge conflicts because I'd already incorporated and released the baseload branch. Generally its easier if you just push bug fixes for a PR to the same branch, rather than creating another branch.
> 
> Can you confirm which of these 3 will need to be merged with master? Or can we close the v2 tooltips? Hard for me to know without reviewing all the changes across all branches.

Think we can close tooltips v2. We need to integrate billing-tooltips-v3 and baseload-v5 into aws-eb-base and eventually into master. billing-tooltips-v3 is based on billing-tooltips-v2, so I am guessing I should have updated billing-tooltips-v2 going forward, and baseload-v5 has the cosmetic fixes based on Claudia's feedback, the only area of shared code should be html_table_formatting.rb - otherwise would have thought there should be no merge conflicts.

Perhaps talk about I update branches as tomorrow's tech meeting, so I can get the policy right to make it easier for you?

PhilipTB:
I have updated the code as the changes to the error reporting in meter_tariff_manager.rb didn't work. So this update is in the current branch rather than a new branch?

ldodds:
Thanks, that was easier to process.

These updates should now be on test.

PR #119, opened 2021-07-16
Restructured alert and data source files into subdirectories

PhilipTB:
- No new code apart from a slight numeric tweak to the targeting and tracking
- Can this be integrated into the front end and deployed on test?
- Have only tested the main functions: adult dashboard, alerts, benchmarks, summary table - but am assuming given it 'compiles' all the rest will work as before
- the baseload breakdown tables are currently broken on test and prod, not sure what is going on as was ok yesterday, but not this morning, can't replicate in the analytics, so suspect an integration/merger issue?
- Reluctant to do any further work on master until this gets 'merged' into master, so in the short term will work on releases derived from this although currently don't have any code development worked planned for the next few days
- says there are conflicts below, says to check out and reproduce, given I did a push then a pull, a 'sync' in VSC terms not sure why it is complaining?


ldodds:
No issues on test server overnight, so am merging this down.

PR #124, opened 2021-07-26
flexible start date, partial months

PhilipTB:
- changed from academic year based start month, to first month of target, or first meter date, or end meter date minus 365 days whichever is the most recent
- added :partial_months to hash being returned for monthly target table


ldodds:
Merged into `aws-eb-test` branch for release to test environment

PR #125, opened 2021-07-28
Fix sign of solar pv co2

PhilipTB:
Fixed sign of solar pv in energy/co2 benchmark, from positive value to negative value. 

The issue manifested itself in the co2 change school comparison. Its a slightly philosophical question how you account for the CO2 from solar PV panels, so ultimately the values may change further, perhaps taking 40g/kWh embedded energy in manufacturing them.

There was is no impact to the co2 on the management summary table as its code had already corrected for the issue, but given this is a fundamental change lower down the code base which should correct in the charts etc., the management summary table adjustment was reversed to take into account the reversal in the underlying data.

Can you please install on test, then integrate to master, then release to production? And notify me when it goes to production as I need to notify the school/Paula who picked up the issue.

### DIFF
--- a/lib/dashboard/aggregation/amr_data.rb
+++ b/lib/dashboard/aggregation/amr_data.rb
@@ -8,6 +8,7 @@ class AMRData < HalfHourlyData
 
   def initialize(type)
     super(type)
+    puts "Got here creating amr_data of type #{type}"
     @total = {}
   end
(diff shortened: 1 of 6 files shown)


ldodds:
Merged into `aws-eb-test` branch for release to test environment

PR #127, opened 2021-08-11
Targeting and tracking fitting

PhilipTB:
new interfaces in target service:
- relevance to be used to cut down list of electricity, gas, SH meters on 'Review targets' page
- enough_data - indication whether t&t will work, with further analytics development this list of school aggregate meters without enough data will dwindle
- annual_kwh_estimate_required? - whether user needs to type in an annual (DEC?) estimate; note if meter data is reset/long gap e.g. B&NES with gas, then the status of this might go from false to true when a gap appears
- culmulative_progress_chart, weekly_progress_chart, weekly_progress_to_date_chart - to be used for the 3 T&T charts used by the front end, so the analytics can the chart definition if necessary, either statically or dynamically
- target_set? - probably not needed by the front end, but used by the analytics testing at the moment as my production YAML unvalidated school data currently doesn;t have many target attributes set, this may not be true for 'test'
- monthly performance versus synthetic version of last year, as alternative to 'variance'

targeting and tracking attributes now work on aggregate meters

Not sure whether this branch is suitable for production, will need to discuss.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ Results/*
 DCC/*
 OldInputData/*
 OldMeterReadings/*
+Simulator/*
 documents/*
 TestResults/*
 BenchmarkLog/*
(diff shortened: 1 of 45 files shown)


ldodds:
@PhilipTB the tests are failing because of a missing gem: awesome_print. Is that a new addition?

PR #137, opened 2021-09-08
Ensure we're returning correct response to enough_data

ldodds:


ldodds:
@PhilipTB this is the proposed change to the service interface. Without this the call to `enough_data_to_set_target?` always returns true, even if an estimate is required. This is the high-level guard that the application uses to ensure that sufficient data is available.

PR #144, opened 2021-09-24
Revised logic, add rspec tests

ldodds:
* Added new method to TargetsService to provide a default start date for fuel type, to be used as suggestion by front-end
* Revised logic in TargetDates class when there is no current target, so that it checks for recent data and also uses a default target start date of either this month or the end date of the amr readings when checking if there's at least a years worth of data
* Added rspec tests for this logic in the TargetDates class

@PhilipTB if you could look this over, that would be helpful. I've added comments to the code and specs.

The application will still decide what the actual start date will be by checking across the various fuel types for the school. I've implemented that logic also.

### DIFF
--- a/app/services/targets_service.rb
+++ b/app/services/targets_service.rb
@@ -36,6 +36,10 @@ class TargetsService
     Holidays.check_school_holidays(@aggregate_school)
   end
 
+  def default_target_start_date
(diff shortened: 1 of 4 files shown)


PhilipTB (commented):
Looks ok on visual inspection

PR #148, opened 2021-10-11
Refactor Sheffield interface into service and api client, add tests

ldodds:
* Create separate PV Live API client, with basic rspec tests
* Create PV Live Service, to hold methods to calculate nearest areas, interpolate and reformat data, delegating api calls to client
* Update test script to refer to new service
* Remove all code


lib/dashboard/data_sources/pv_live/pv_live_api.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def get_data(path, params={}, headers={})
+      response = Faraday.get(BASE_URL + path, params, headers)
+      handle_response(response)
+    end
+
+    def handle_response(response)
+      raise ApiFailure.new(response.body) unless response.success?
+      body = JSON.parse(response.body, symbolize_names: true)
+      raise ApiFailure.new(body[:error_description]) if body[:error_description]
```
jhigman, line 34: The errors raised here will be immediately handled in the rescue for the method - is that intended, or is the rescue in the method only supposed to handle other error conditions? (I had something similar in the geo api...)

jhigman (approved):

PR #153, opened 2021-10-21
Brighton solar edge bulk download enhancements

PhilipTB:
Update to test script only, just needs including in master, thanks


ldodds:
This has changes to both the meter class (minor) and the meter breakdown advice (looks like fairly substantial additions) so its more than just changes to test script. I'll merge but not release immediately.

PR #185, opened 2022-02-03
1st version

PhilipTB:
Should be integrated into analytics master as standalone/independent of other code for moment, so safe


ldodds (changes requested):
Need to avoid holding references to the aggregate school and meter data, if this is going to be cached.

ldodds (approved):

PR #204, opened 2022-02-16
temp fix to referencing problem from front end

PhilipTB:


lib/dashboard/utilities/record_test_times_to_csv.rb
```diff
@@ -67,6 +67,12 @@ def save_summary_stats_to_csv
 
   private
 
+  def default_test_directory
```
ldodds, line 70: This doesn't work because its relying on a call to `rails?` to decide whether to create the directory. But `rails?` relies on the `@rails` member variable which hasn't been defined at this point in the code.

ldodds (changes requested):

lib/dashboard/utilities/record_test_times_to_csv.rb
PhilipTB: would changing:
  def default_test_directory
    return nil if rails?

    TestDirectory.instance.timing_directory
  end

to

  def default_test_directory
    return nil if Object.const_defined?('Rails')

    TestDirectory.instance.timing_directory
  end

work?

Reluctant to follow method wrapping or alias methods, as I think it implies doing this around every timing call?

Can we have a Meet call at midday? Invite pending

ldodds:
I've added a check for the rails constant so we can merge this and deploy all the other fixes, plus the community use code onto test.

ldodds (approved):

PR #214, opened 2022-03-14
Check whether we can generate synthetic data without a target

ldodds:
Use the default target start date (this month) if we don't have a target.


ldodds:
@PhilipTB revised logic for that method and comments as requested. With the revised code I can now set targets in the application for a school that has an annual estimate.

PR #216, opened 2022-03-15
Update targeting and tracking tests

PhilipTB:
Would be useful if this branch made its way to master as it has better self checking of dates combinations the analytics can't handle and will save my time in future tracking down issues - I have chased this issue twice, hence the specific check and Exception thrown for poor user configuration of T&T dates

### DIFF
--- a/app/models/meter_collection.rb
+++ b/app/models/meter_collection.rb
@@ -488,7 +488,11 @@ class MeterCollection
   end
 
   def process_school_times(school_day_times, community_times)
-    @open_close_times = OpenCloseTimes.convert_frontend_times(school_day_times, community_times, holidays)
+    if school_day_times.nil? # TODO(PH, 17Feb2022) remove once new school timing code has migrated to production, backwards compatibility with old YAML files
+      @open_close_times = OpenCloseTimes.default_school_open_close_times(holidays)
+    else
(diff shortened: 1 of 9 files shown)


ldodds:
@PhilipTB what are the new date checks in this branch? Apart from changes to your test framework, I can see some refactoring of the open/close times and a new exception being thrown from the solar analysis. Am wondering what date combinations are being tested/handled here.

PhilipTB:
> @PhilipTB what are the new date checks in this branch? Apart from changes to your test framework, I can see some refactoring of the open/close times and a new exception being thrown from the solar analysis. Am wondering what date combinations are being tested/handled here.

I think from recollection it was the circumstance where the first target date was before the 1st actual meter reading, something the analytics currently doesn't support and which has happened twice (Wimbledon being the second example) and which took me an hour or so to debug and then realised I had seen the problem before and it was probably better if I checked and threw an explicit exception.

ldodds:
Can you confirm that the expected changes are in this PR? The codes changes are:

- some changes for your test harness, include a test for Wimbledon
- catching an exception from statssample
- a refactoring of how OpenCloseTimes is initialised

There's nothing relating to target start dates. And we don't want an exception for that, as in the revised approach it will be legal to do that. 

Am really just checking all the expect changes are in the PR.

PR #247, opened 2022-05-04
Reduced test script memory usage

PhilipTB:
Stopes paging and reduces memory usage for all_tests and all schools below 700MB. There is potentially more to be done, potential to keep usage below 300MB but would require changes to the application.


ldodds:
@PhilipTB looking at the code there seems to be a mixture of functional updates to some of the alert code AND changes to the test framework. Should those both be in this branch?

PR #258, opened 2022-05-12
Only build school map if its not already built

ldodds:
The Benchmark Manager builds a map of schools used to help with formatting links to schools.

At the moment this map is rebuilt every time that a table is run (which is once for the chart, and then once for every variation of the table). 

This PR tweaks the code so that the school map is only built once. The list of schools and data doesn't change between calls, so this is safe to do.


PhilipTB (approved):
Looks ok.

PR #261, opened 2022-05-24
release

PhilipTB:
- fix thermostatic scatter charts in front end
- older analytics tests upgraded to recent test infrastructure(model fitting, community use)
- St Julians High - support real solar generation data but calculate export and self consumption data synthetically
- add more debug to solar PV aggregation
- However the last day of St Julians High mains consumption data is incorrect - fed as all zeros


ldodds:
Rebased this with master to fix up the build failure which was caused by a security vulnerability warning in nokogiri.

PR #277, opened 2022-07-12
Remove Mechanize

ldodds:
It doesn't look like we use the Mechanize gem at all. We occasionally have to upgrade it due to bugs with it and nokogiri. But if we don't use it, then no sense keeping it as a dependency.

There's no direct references to it in any of the application, analytics code (including the scripts) or statsample. It was added to the project around 2018 so may have been used in early work?

This PR removes Mechanize. As a result, its dependencies are removed too which includes a number of other gems. I've not checked to see if these might be used elsewhere.

@PhilipTB are you aware of Mechanize being used in the past? Its for performing web requests, working with web pages, etc.


jhigman (approved):

ldodds:
Further review shows this was added [in a commit in Jun 2018](https://github.com/Energy-Sparks/energy-sparks_analytics/tree/6d8fb72071bd6080860325fe70701981179788cc) when the first Gemfile was added. While Mechanize was added to the Gemfile, it doesn't seem to have been used in the scripts which at that point did direct includes.

So I think this may have been added as something to be used later and has been there ever since.

PR #286, opened 2022-07-25
Nurserys with no holidays support

PhilipTB:


ldodds:
I'm holding off on merging this initial prototype code until we've made a decision about what to do with supporting nurseries.

PR #295, opened 2022-08-03
AWS 3.4.8 Upgrades

ldodds:
Upgrade Ruby version to 2.7.6 (`2.7.6p219` is the supported version on AWS) and run bundler

Using rvm I had to:

```
rvm install 2.7.6
rvm use 2.7.6 #or just navigate to the directory to trigger rvm to load the right ruby
gem install bundler:2.2.32
bundle
```


ldodds:
@PhilipTB just a heads-up that we're planning on merging this down tomorrow to do the test server upgrade ahead of live next week. So you will likely need to update to ruby 2.7.6 and reinstall gems.

ldodds:
@PhilipTB ignore previous message, i'll delay these updates until next week so we can get your other changes released first.

iant (approved):

PR #304, opened 2022-08-16
Localise variables produced by baseload alerts

ldodds:
* Extracts the `String` variables produced by the baseload alerts from the code into YAML files, all the others are largely handled by the `FormatEnergyUnit` class which will be further localised separately
* Adds a base class helper to `ContentBase` to help to standardised key names across the alert classes (have settled on `underscore` version of class name). This is potentially brittle if classes get renamed, but most of these classes can't be renamed without database updates in the application. This approach still allows the files to be reorganised. The alternative was to declare a constant on each class which felt redundant
* Changed code to lookup relevant variable text via the `I18n.t` helper. This is a mixture of words and short sentences
* Adds an `I18nHelper` class to support look up of common helper text, e.g adjectives and dates
* With the text extracted to the YAML files, it was possible to refactor the code to move the `timescale` method to `AlertBaseloadBase` avoiding need for every class to have it

I think we're going to inevitably have some words/phrases duplicated across the analytics and application, e.g. day names, etc. There's no way to share YAML both ways. I think this is probably acceptable as its a small amount of text.

The approach I'm taking here is to tackle a directory of alerts at a time. Checking the `TEMPLATE_VARIABLES` for each class to find any that are `Strings`. These generally need specially handling while the rest get formatted via `FormatEnergyUnit`.

To test the changes I've been running the `scripts/test_alerts.rb` script. I first ran this for a couple of schools without any changes, and am then re-running after refactoring. The script either produces errors or reports on any differences in the text.

So while there are no real rspec tests, there has been some manual testing.

### DIFF
--- a/lib/dashboard/alerts/electricity/baseload/alert_electricity_baseload_versus_benchmark.rb
+++ b/lib/dashboard/alerts/electricity/baseload/alert_electricity_baseload_versus_benchmark.rb
@@ -170,14 +170,10 @@ class AlertElectricityBaseloadVersusBenchmark < AlertBaseloadBase
     :alert_1_year_baseload
   end
 
-  def timescale
-    'last year'
-  end
-
   def commentary
     [ { type: :html,  content: evaluation_html } ]
   end
-  
+
   def evaluation_html
     text = %(
               <% if average_baseload_last_year_kw < benchmark_per_pupil_kw %>
@@ -185,7 +181,7 @@ class AlertElectricityBaseloadVersusBenchmark < AlertBaseloadBase
                 <%= format_kw(average_baseload_last_year_kw) %> compared with a
                 well managed school of a similar size's
                 <%= format_kw(benchmark_per_pupil_kw) %> and
-                an examplar schools's 
+                an examplar schools's
                 <%= FormatEnergyUnit.format(:kw, @exemplar_per_pupil_kw) %>,
                 but there should still be opportunities to improve further.
               <% else %>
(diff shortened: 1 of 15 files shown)


lib/dashboard/alerts/electricity/baseload/alert_electricity_baseload_versus_benchmark.rb
```diff
@@ -268,10 +264,9 @@ def enough_data
 
   def summary_text
     if @one_year_saving_versus_exemplar_£ > 0
-      'Your baseload is high, reducing it could save ' +
-      FormatEnergyUnit.format(:£, @one_year_saving_versus_exemplar_£, :text) + 'pa'
+      I18n.t("#{i18n_prefix}.summary.high", saving: FormatEnergyUnit.format(:£, @one_year_saving_versus_exemplar_£, :text))
     else
-      'You are doing well - you are an exemplar school'
+      I18n.t("#{i18n_prefix}.summary.ok")
```
iant, line 269: may not be necessary but maybe we can add a fallback to original text here? (I'm doing similar in the other pr)
`I18n.t("#{i18n_prefix}.summary.ok", default: nil) || 'You are doing well - you are an exemplar school'`

lib/dashboard/alerts/electricity/baseload/alert_electricity_baseload_versus_benchmark.rb
```diff
@@ -226,8 +222,8 @@ def enough_data
 
     @one_year_saving_versus_benchmark_kwh = @average_baseload_last_year_kwh - @one_year_benchmark_by_pupil_kwh
     @one_year_saving_versus_benchmark_£   = @one_year_saving_versus_benchmark_kwh * electricity_tariff
-    @one_year_saving_versus_benchmark_co2 = @one_year_saving_versus_benchmark_kwh * blended_co2_per_kwh  
-    @one_year_saving_versus_benchmark_adjective = @one_year_saving_versus_benchmark_kwh > 0.0 ? 'higher' : 'lower'
+    @one_year_saving_versus_benchmark_co2 = @one_year_saving_versus_benchmark_kwh * blended_co2_per_kwh
+    @one_year_saving_versus_benchmark_adjective = @one_year_saving_versus_benchmark_kwh > 0.0 ? I18nHelper.adjective('higher') : I18nHelper.adjective('lower')
```
iant, line 226: as it's used twice here and below maybe this could be extracted to a method on the i18n helper?  something like...
`@one_year_saving_versus_benchmark_adjective = I18nHelper.adjective_for(@one_year_saving_versus_benchmark_kwh)`


```
def adjective_for(result, greater_than_value = 0)
   return I18nHelper.adjective('higher') if result > greater_than_value
   
   I18nHelper.adjective('lower')
end
```

lib/dashboard/alerts/electricity/baseload/alert_electricity_baseload_versus_benchmark.rb
ldodds: I'm removing the defaults for now, it helps with my testing that I know where its coming from. We could do the same with charts eventually.

PR #306, opened 2022-08-17
Setup up github action to run tests

iant:


iant:
hi @ldodds a few bits on this pr...

this is running the standard test and seemingly saves all artifacts ok.  However, it errors when `RunBenchmarks` is uncommented (see diff) with a not very helpful error message.  It may be timing out as it runs fine locally with the uploaded data file.  any thoughts? 

I've also added a couple of the other tests to see if they run ok too - some do, some don't.

I also needed to comment out the logger. 

It also runs all tests in the spec folder which I think may be better moved to a separate github action?

iant:
^^ hi @ldodds update to the above.  I've added a new script copied from standard_test.rb. I removed the logging (which was causing errors) and added a single check for a specifically title and anonymised meter reading yml file.  I think we'll need to revisit why the benchmark run is erroring (could be a timeout, could be some logging issue) but this is most of the way there. 

There are now 2 separate jobs defined in the github action: 1 for the rspec tests and 1 for the analytics tests.

ldodds:
@iant this is an excellent start. There's some separate issues with logging and possibly benchmarks, which we can resolve separately. I couldn't quickly see the error messages, do you have

One bit of feedback: to keep artefact sizes smaller, is there a way to exclude the contents of the MeterCollections directory (so the YAML and marshall file) from the zip/upload?

Otherwise I'm happy for this to be merged down, and we can improve over time.

iant:
hi @ldodds 
> One bit of feedback: to keep artefact sizes smaller, is there a way to exclude the contents of the MeterCollections directory (so the YAML and marshall file) from the zip/upload?

^^ Looks like we can we can ignore by pattern 
https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions 
I've just pushed an update.  Would you be able to double check the artifacts remaining are all as expected (excluding the MeterData) and if all good I'll get this merged.

ldodds (approved):
Looks good to me. This allows us to move away from travis and gives a starting point for automating runs of the analytics.

PR #316, opened 2022-08-19
Add a rake task to create some test data

iant:


rakelib/testing.rake
```diff
@@ -74,4 +74,46 @@ namespace :testing do
...
+        end
+      end
+    end
+
+    $stderr.puts "Anonymising files"
+    schools.each_with_index do |school, index|
+      file_name = "./#{test_dir}/MeterCollections/unvalidated-data-acme-#{index}.yaml"
+      text = File.read(file_name)
+      text = text.gsub(/.*  :id:.*/, "  :id: 00000#{index+1}")
+      # text = text.gsub(/.*  :name:.*/, "  :name: Acme School #{index}")
```
ldodds, line 112: Should this be commented out? I think we should be renaming the schools too?

There's also the `location` attribute to remove the lat/lng

ldodds (changes requested):

ldodds (approved):

PR #317, opened 2022-08-19
fix for noisy solar PV export schools

PhilipTB:


lib/dashboard/charting_and_reports/charts/chart_configuration.rb
```diff
@@ -648,6 +648,12 @@ def self.standard_series_label_substitution(type, school_type)
         ]
       },
     },
+    solar_pv_single_day_adhoc_test: {
```
ldodds, line 651: Charts need to be defined in the chart configuration YAML, and there's tests now in master to confirm that has been done. So merging this will end up breaking the build.

Can you either add this chart to the YAML file, or remove it if it was just for ad hoc testing?

I'd suggest merging in latest from master as the YAML files have been tidied.

ldodds (changes requested):

lib/dashboard/charting_and_reports/charts/chart_configuration.rb
PhilipTB: Leigh,

I have done my best to make this change as requested. Its a little bit difficult at the moment as I have no broadband, relying on slow 3G mobile and I haven't upgraded to Ruby 2.7.6 and done a bundle update  recently, partly because of the broadband issue and partly as I am trying to limit my hours until the end of the month. Was planning on doing the 2.7.6/bundle update early September.

When I ran rspec, now getting lots of statsample errors e.g.:

> An error occurred while loading ./spec/lib/dashboard/x48_formatter_spec.rb.
> Failure/Error: require 'statsample'
> 

> LoadError:
>   cannot load such file -- statsample
> # ./lib/dashboard.rb:42:in `<top (required)>'
> # ./spec/spec_helper.rb:27:in `<top (required)>'
> # ./spec/lib/dashboard/x48_formatter_spec.rb:1:in `<top (required)>'

And the 'i18n-tasks normalise' you suggested I call didn't work:

> C:\Users\phili\OneDrive\ESDev\energy-sparks_analytics>i18n-tasks normalise
> warning: parser/current is loading parser/ruby27, which recognizes2.7.6-compliant syntax, but you are running 2.7.5.
> Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
> i18n-tasks: unknown command: normalise

However, 'i18n-tasks normalize', may have worked:

> C:\Users\phili\OneDrive\ESDev\energy-sparks_analytics>i18n-tasks normalize
> warning: parser/current is loading parser/ruby27, which recognizes2.7.6-compliant syntax, but you are running 2.7.5.
> Please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
> #StandWithUkraine

However, not sure about the '#StandWithUkraine' tag which came back?

Philip

ldodds (approved):

PR #322, opened 2022-08-24
I18n Gas & storage heater alerts

ldodds:
Localise the strings produced as template variables from the gas and storage heater alerts.

Involves:

* replacing member variables with methods, so they can be called with a different locale
* turning ERB templates into I18n.t calls
* adding a couple of specs for helper methods that were returning strings and ratings,

Also adds comments to a couple of alerts that are no longer needed.

Dev checklist.

* [x] storage heaters
* [x] main gas alerts
* [x] hot water
* [x] boiler control

### DIFF
--- a/config/locales/alerts/gas/boilercontrol.yml
+++ b/config/locales/alerts/gas/boilercontrol.yml
@@ -16,5 +16,5 @@ en:
       timescale: this holiday
     alert_heating_sensitivity_advice:
       timescale: year
-    alert_heating_on_non_school_days:
-      timescale: 1 year
+    alert_optimum_start_analysis:
+      timescale: last year
--- a/config/locales/alerts/gas/boilercontrol.yml
+++ b/config/locales/alerts/gas/boilercontrol.yml
@@ -1,15 +1,16 @@
 ---
 en:
   analytics:
+    alert_turn_heating_off:
+      timescale: next week
     alert_heating_coming_on_too_early:
       timescale: 7 days
     alert_heating_hot_water_on_during_holiday_base:
-      timescale: this holiday
+      both: the hot water has been left on on %{hotwater_days} days and the heating on %{heating_days} days
       heating_not_on: Well done you have used no gas this holiday.
+      only_heating: the heating has been left on on %{heating_days} days
+      only_hotwater: the hot water has been left on on %{hotwater_days} days
       summary: |
         Your %{heating_type} has been left on over the %{holiday_name} holiday. Up
         until %{date} %{left_on_message} costing %{cost}, and a projected %{project_cost} by the end of the holiday.
-      only_hotwater: "the hot water has been left on on %{hotwater_days} days"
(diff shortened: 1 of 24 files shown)


lib/dashboard/alerts/gas/boiler control/alert_heating_hotwater_on_during_holiday.rb
```diff
@@ -138,31 +140,35 @@ def calculate(asof_date)
   end
   alias_method :analyse_private, :calculate
 
+  def holiday_name
+    @holiday_period.nil? ? 'Not a holiday' : @holiday_period.title
```
iant, line 144: can 'Not a holiday' be moved to a translation file?

lib/dashboard/alerts/gas/boiler control/alert_heating_off.rb
```diff
@@ -96,16 +96,20 @@ def self.template_variables
       description: 'days since last meter reading date',
       units:  Integer
     },
+    days_ago: {
+      description: 'days since last meter read, including day/days',
+      unit: String
```
iant, line 101: can this text be moved to a translation file?  (tho looks like there's quite a bit of text here to extract out here so might be better in a future pr)

config/locales/alerts/gas/boilercontrol.yml
```diff
@@ -0,0 +1,24 @@
+---
+en:
+  analytics:
+    alert_heating_coming_on_too_early:
+      timescale: 7 days
+    alert_heating_hot_water_on_during_holiday_base:
+      both: the hot water has been left on on %{hotwater_days} days and the heating on %{heating_days} days
```
iant, line 7: does this need separate `one:` and `other:` keys (passing a `count`)?  (e.g. "The heating has been left on 1 day") 
same for `only_heating:` and `only_hotwater:`?

iant (approved):
hi @ldodds all looks good to me.  couple of v minor comments on translated text.

lib/dashboard/alerts/gas/boiler control/alert_heating_off.rb
ldodds: The descriptions are only used in the admin view, so not planning to translate any of that.

lib/dashboard/alerts/gas/boiler control/alert_heating_hotwater_on_during_holiday.rb
ldodds: I didn't translate that because this is basically a fallback that will only be seen by admin users, e.g. if we're checking why an alert wasn't generated. If the holiday period is nil then the alert isn't suitable for display to users.

So will leave this for now.

config/locales/alerts/gas/boilercontrol.yml
ldodds: Yes, good catch, have adjusted the format and code.

PR #323, opened 2022-08-24
Localise chart series data

iant:


iant:
hi @ldodds this can be merged now i think - the rest of the series work will be handled in this (still wip, but close) pr https://github.com/Energy-Sparks/energy-sparks/pull/1971 

also fyi community and community baseload series support has been added

config/locales/common.yml
```diff
@@ -43,3 +43,44 @@ en:
       primary: Primary
       secondary: Secondary
       special: Special
+    series_data_manager:
```
ldodds, line 46: Might be better to just have `analytic.series_name` as the key, rather than nesting it under `analytics.series_data_manager.series_name`?

ldodds (approved):
Looks fine, but brief comment on the `series_name` key, up to you if you want to change.

PR #324, opened 2022-08-24
I18n time comparison alerts

ldodds:
Localise the strings produced as template variables from the time comparison alerts:

Involves:

* replacing member variables with methods, so they can be called with a different locale
* adding an additional helper for adjectives
* reworking some of the method inheritance around generating names for periods to make it easier to localise
* removing custom ordinal code cribbed from ActiveSupport and instead rely on `.ordinalize` to support hooking in custom code
* also fixes a bug where being unable to find current holiday triggers an error, it now cleanly falls back to indicating there is not enough data

### DIFF
--- a/config/locales/alerts/time_period_comparison/common.yml
+++ b/config/locales/alerts/time_period_comparison/common.yml
@@ -1,4 +1,0 @@
----
-en:
-  analytics:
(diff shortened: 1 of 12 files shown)


iant (approved):
hi @ldodds all looks good to me 👍

PR #326, opened 2022-08-25
Gemfile.lock for Windows; Benchmark refactoring

PhilipTB:


ldodds:
This PR won't fix the issue we're trying to resolve as the problem is in how the benchmark results are being round-tripped through the database: `Float::INFINITY` is being turned into nil when serialised as JSON whereas in the YAML version its round-tripped.

With that in mind, is there any value in still merging this PR? There are changes to the Gemfile.lock which we'll need to add in at some point I think. The other changes were debug code, plus some changes to how columns are filtered out.

PhilipTB:
> This PR won't fix the issue we're trying to resolve as the problem is in how the benchmark results are being round-tripped through the database: `Float::INFINITY` is being turned into nil when serialised as JSON whereas in the YAML version its round-tripped.
> 
> With that in mind, is there any value in still merging this PR? There are changes to the Gemfile.lock which we'll need to add in at some point I think. The other changes were debug code, plus some changes to how columns are filtered out.

I think there is no harm in merging because its at least as good as what is out there at the moment.

I had forgotten about the Gemfile.lock, problem is I suspect if I revert my environment is going to complain and then I will have to do a bundle update which will automatically update the Gemfile.lock - are my (automatic) updates causing a problem?

ldodds (approved):

ldodds:
It looks like we previously had Windows versions of the gems in the Gemfile.lock, these were likely automatically removed after the upgrade.

I haven't had any issues, but will see when this is properly merged and added to the app.

Updated title of the PR to reflect changes.

PR #329, opened 2022-08-31
Rework localisation of :day charts

iant:
Charts with `x_axis: :day,` either rely on a default date format (maybe `Date.to_json`) or an explicit date format given in the `x_axis_reformat` configuration.  This pr: 

- [x] reverts change to XBucketDay, so it generates Dates and not string (this will fix trendline)
- [x] ensures that charts formatted via `x_axis_reformat` are localised
- [x] ensures that all charts that are `:day` have an explicit `x_axis_reformat`
- [x] adds a test to validate that any chart with a :day x-axis also has a format


PhilipTB:
Ian, I left a more detailed comment against 1 of the commits, you shouldn't need to edit any rb files ending _deprecated as they shouldn't be called, and the reformat function which is currently in use is in AggregatorPostProcess#reformat_x_axis which currently hasn't been localised. Also its not clear to me why all x_axis: : day chart configs now need a reformat config? Philip

PhilipTB:
I also feel you need to change:

  def create_trend_lines
    calculate_regression_parameters_outside_model

    if Object.const_defined?('Rails')
      add_trendlines_for_rails_all_points
    elsif false && Object.const_defined?('Rails')
      add_trendlines_for_rails_2_points
    else
      analytics_excel_trendlines
    end 
  end

to

  def create_trend_lines
    calculate_regression_parameters_outside_model

    if false && Object.const_defined?('Rails')
      add_trendlines_for_rails_all_points
    elsif Object.const_defined?('Rails')
      add_trendlines_for_rails_2_points
    else
      analytics_excel_trendlines
    end 
  end

Should reduce the payload of data being sent to the front end for trendlines?

iant:
thanks @PhilipTB good spot - I've updated the correct `reformat_x_axis` method in the pr.
I'll catch up with Leigh regarding the `create_trend_lines` we may need to do that in a separate pr.   

hi @ldodds can we have a quick chat before you review this pr?

lib/dashboard/charting_and_reports/charts/chart_configuration.rb
```diff
@@ -2242,4 +2254,41 @@ def self.standard_series_label_substitution(type, school_type)
...
+    optional(:x_axis_reformat).value(:hash)
+    optional(:meter_definition).value(:symbol)
+    optional(:drilldown_name).value(array[:string])
+    optional(:chart1_type).value(:symbol)
+  end
+
+  rule(:x_axis_reformat) do
+    x_axis_reformat = values[:x_axis_reformat]
+
+    if !x_axis_reformat.nil? && !x_axis_reformat&.key?(:date)
```
iant, line 2274: `if x_axis_reformat.present? && x_axis_reformat.is_a?(Hash) && !x_axis_reformat&.key?(:date)`

add extra check for anything other than a hash

lib/dashboard/charting_and_reports/charts/chart_configuration.rb
```diff
@@ -2242,4 +2254,41 @@ def self.standard_series_label_substitution(type, school_type)
...
+  end
+
+  rule(:x_axis) do
+    x_axis, x_axis_reformat = values[:x_axis], values[:x_axis_reformat]
+
+    key.failure("x_axis_reformat needs to be present if x_axis is a day symbol") if x_axis == :day && x_axis_reformat.nil?
+  end
+
+  rule(:name) do
+    name, inherits_from = values[:name], values[:inherits_from]
```
iant, line 2290: remove

lib/dashboard/charting_and_reports/charts/chart_configuration.rb
```diff
@@ -120,6 +122,7 @@ def self.standard_series_label_substitution(type, school_type)
       chart1_subtype:   :stacked,
       meter_definition: :all,
       x_axis:           :year,
+      x_axis_reformat:  { date: '%A %-d %b %Y' },
```
iant, line 125: remove this line

iant:
closing this for now.  as discussed we'll pick up the validations in a future bit of work...

PR #338, opened 2022-09-07
fix

PhilipTB:
Ian there was a mistake in the YAML, could you include this PR in the master release you have just done?


iant (approved):
hi @PhilipTB thanks :+1: no probs at all

PR #339, opened 2022-09-23
Unhandled x_axis drilldown config

iant:
We've been intermittently getting this error over the past few weeks.

See Rollbar:

https://rollbar.com/energysparks/EnergySparksProduction/items/1407/?utm_campaign=period_summary&utm_medium=email&utm_source=rollbar-notification&utm_content=show_recent_and_monthly_usage

https://rollbar.com/energysparks/EnergySparksProduction/items/1406/?item_page=0&item_count=100&#instances

https://rollbar.com/energysparks/EnergySparksProduction/items/1408/?item_page=0&item_count=100&#traceback

Most recently for "Castlefield Campus" who only completed onboarding on Friday.

Its possible this is a data related issue with determining the drill-down config.

### DIFF
--- a/lib/dashboard/charting_and_reports/charts/chart_drilldown.rb
+++ b/lib/dashboard/charting_and_reports/charts/chart_drilldown.rb
@@ -97,6 +97,8 @@ class ChartManager
   end
 
   def x_axis_drilldown(existing_x_axis_config)
+    return unless existing_x_axis_config
+
     case existing_x_axis_config
     when :year, :academicyear
       :week


lib/dashboard/charting_and_reports/charts/chart_drilldown.rb
```diff
@@ -97,6 +97,8 @@ def drilldown_available(chart_config_original)
   end
 
   def x_axis_drilldown(existing_x_axis_config)
+    return unless existing_x_axis_config
```
iant, line 100: hi @PhilipTB i was wondering if you might be able to help me with this please when you get a mo?  

there's a bunch of errors, seemingly for a newly set up school, where a `nil` `existing_x_axis_config` is raising an error. (rollbar links in the pr description).  this is as expected in the case statement in the `x_axis_drilldown` method (`raise EnergySparksBadChartSpecification...`).   

As it's occuring when `drilldown_available?` is called, I think it's ok to return `nil` here (so `drilldown_available` returns `false` - ie. if there's no config then no drilldown is available). 

However, I was wondering if there's something upstream of this that's resulting in a `nil` `existing_x_axis_config` being assigned in the first place and if this would need to be debugged?

thanks 🙏   (happy to jump on a call if none of the above is clear)

PhilipTB: Ian,

Generally I would want to avoid this type of fix if it can be fixed further upstream where the root cause of the problem is, a fix of this type may cause other problems. Ultimately the question is is why is the x-axis config nil, its not an expected config as far as I can tell. Can you reproduce the error? I have tried drilldown on several charts for the school and it all seems to work without crashing.
And looking at the school I have noticed 3 other issues for which Trello cars need raising:
1. the management dashboard table displays an artifact of internal defaulting of dates for no date data: "Electricity data: 1 Jan 4000 - 1 Jan 1000. Gas data: 1 Aug 2021 - 1 Nov 2021. More information" which needs changing
2. the frost protection page is showing but without a chart: https://energysparks.uk/schools/castlefield-campus/analysis/2510217
3. the daily variation in use https://energysparks.uk/schools/castlefield-campus/analysis/2510220 and cost pages https://energysparks.uk/schools/castlefield-campus/analysis/2510219 for electricity are showing when they shouldn't
I am happy to raise Trello cards for these if you like? Its worth taking a copy of the school's YAML today before Becca fixes the data for the school (I have just done so) and afterwhich we then would find it difficult to reproduce.

However, ultimately all of the above is moot in that the school is not 'data visible' and it shouldn't be made visible until suitable data is available, the gas data for example is limited and a year out of date and there is clearly no electricity data.

Overall, I think we need to understand why the drilldown is throwing an error by reproducing it, I suspect the highlevel, original chart shouldn't be being displayed. Then suspect if we do actually need to fix the  drilldown method which is unlikely  then we should raise a 'EnergySparksnotEnoughDataException' in this circumstance?

iant: Sounds like we may need something in place to prevent chart requests for non 'data visible' schools.  I'll follow up with @ldodds next week.  thanks @PhilipTB -- much appreciated.

ldodds:
Closed this because, as the Rollbar log shows, the error is in running the dashboard and analysis page charts are part of the content batch. Its not as a result of trying to display charts.

The school has nearly two years of electricity data, although its lagging behind for one of the meters. However, post aggregation, the aggregate electricity meter has no data in it. This is causing the chart to fail, default dates and incorrect values on the dashboard and probably the other issues. 

Will put further updates on the trello card.

PR #344, opened 2022-10-26
Fix for heating during school holidays alert

iant:


lib/dashboard/alerts/gas/boiler control/alert_heating_hotwater_on_during_holiday.rb
```diff
@@ -124,6 +124,12 @@ def calculate(asof_date)
       @heating_days_so_far_this_holiday  = count_days(boiler_usage, :heating)
       @hotwater_days_so_far_this_holiday = count_days(boiler_usage, :hotwater)
 
+      @relevance = if @holiday_usage_to_date_kwh.zero?
```
ldodds, line 127: Perhaps end with a conditional that sets both @relevant and @rating, e.g. either relevant and rating of zero, or not_relevant a rating of 10 (to match if @school.holidays.holiday?(asof_date) was false)

ldodds (changes requested):

lib/dashboard/alerts/gas/boiler control/alert_heating_hotwater_on_during_holiday.rb
iant: :+1: done

ldodds (approved):

PR #348, opened 2022-10-28
Fix for Annual Use advice page

iant:


iant:
@ldodds `AdviceGasAnnual` inherits from `AdviceGasBase` (so no `valid_meters` afaics) but i think this pr will fix the issue for `AdviceElectricityAnnual`

ldodds:
@iant yes that does look like it fixes the electricity page.

I'd missed that `AdviceGasAnnual` extends `AdviceGasBase`, rather than `AdviceBenchmark`. 

However it implements `normalised_benchmark_chart_name` which is a method only called by `AdviceBenchmark`. So I'm now wondering if there's also a bug in `AdviceGasAnnual` and it *should* extend that class. In which case the `valid_meters` check may also need changing. 

That method was added [in this commit](https://github.com/Energy-Sparks/energy-sparks_analytics/commit/2ecba8095dfc6710fa2ed675a3fbf51255bec935#diff-cb769122410e99b2b318eedc770ab91fddd6993722d9e759b9deec3d0885e755) where a few advice classes were split out from `advice_general.rb`

The commit also included changes to `AdviceStorageHeaters` which adds the same `normalised_benchmark_chart_name` method. So it also looks like it was possibly intended to extend `AdviceBenchmarks`?

@PhilipTB can you clarify here? Either 2 of the advice classes should be extending AdviceBenchmark (and may need further changes) or the `normalised_benchmark_chart_name` methods can be removed as they're never going to be called.

ldodds:
@iant after discussion with @PhilipTB it looks like both `AdviceStorageHeaters` and `AdviceGasAnnual` should extend `AdviceBenchmark`. They should both trigger the inclusion of an extra chart for normalised usage if the school floor area or pupil count has changed. At the moment, because of the inheritance and configuration, they're both always including those charts, rather than when its useful to do so.

Can you:

* update both classes to extend `AdviceBenchmark`
* review whether `valid_meters` and any of the other methods that relate to meter types need to be updated
* test whether the pages still generate correctly for schools with gas, electricity and storage heaters.

`nottingham-girls-high-school` has both gas and electricity for testing the basic functionality
`mallaig-high-school` has storage heaters.

It doesn't look like there are any schools configured with varying floor areas yet, so for now we just need to ensure that the pages still generate. For the above schools the pages should end up with one less chart.

iant:
:+1: thanks @ldodds & @PhilipTB!  I'll make those changes this morning.

iant:
@ldodds @PhilipTB would you be able to give this another quick review when you get a mo please.  specifically if the correct valid_meters have been used for each (I used the meter collections defined here https://github.com/Energy-Sparks/energy-sparks_analytics/blob/master/app/models/meter_collection.rb#L100)

ldodds:
@iant the `valid_meter` methods on each class should be consistent with what is being returned in their `aggregate_meter` method. This ensures that any data checks will be using the same meter(s) as used for the calculations.

The `relevance` checks use the `.electricity?`, `.gas?`, etc helpers on `MeterCollection` these just wrap calls to, e.g `aggregated_electricity_meters`. 

So as long as all three methods are consistent, then you're OK.

lib/dashboard/advice/advice_storage_heaters.rb
```diff
@@ -14,4 +14,10 @@ def aggregate_meter
   def normalised_benchmark_chart_name
     :benchmark_storage_heater_only_£_varying_floor_area_pupils
   end
+
+  private
+
+  def valid_meters
+    [@school.storage_heater_meter]
```
iant, line 21: hi @PhilipTB can i just double check with you that this is the correct "valid meter" to use for storage heaters?

PhilipTB: yes, looks ok

iant: thanks @PhilipTB 

@ldodds i think this is ready for a last review.  i'll do some testing with the 2 mentioned schools.

ldodds (approved):
OK to merge and then incorporate this into the application.

Suggest releasing it tomorrow and checking that pages are displaying as expected. We can release to production next week.

iant:
@ldodds will merge this today.  all ok to still do a tagged release of this on test?

ldodds:
Yes OK to merge @iant. As discussed, to get these onto test please do an analytics only PR then merge that into aws-eb-test so we can review on the test server before it gets merged down to master.

PR #350, opened 2022-10-30
Economic costs change over time

PhilipTB:
New meter attribute economic_tariff_change_over_time allows economic tariffs to change over time; previous economic_tariff meter attribute supported for backwards compatibility.

### DIFF
--- a/app/models/meter_attributes.rb
+++ b/app/models/meter_attributes.rb
@@ -457,6 +457,28 @@ class MeterAttributes
     )
   end
 
+  def self.default_economic_rates
+    MeterAttributeTypes::Hash.define(
+      required: true,
+      structure: {
+        per:  MeterAttributeTypes::Symbol.define(required: true, allowed_values: [:kwh]),
+        rate: MeterAttributeTypes::Float.define(required: true)
+      }
+    )
+  end
+
+  def self.default_economic_rates_by_time_of_day
+    MeterAttributeTypes::Hash.define(
+      required: false,
+      structure: {
+        per:  MeterAttributeTypes::Symbol.define(allowed_values: [:kwh]),
+        rate: MeterAttributeTypes::Float.define,
+        from: MeterAttributeTypes::TimeOfDay.define,
+        to:   MeterAttributeTypes::TimeOfDay.define
+      }
+    )
+  end
+
   class EconomicTariff < MeterAttributeTypes::AttributeBase
     id :economic_tariff
     key :economic_tariff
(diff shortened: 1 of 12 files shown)


app/models/meter_attributes.rb
```diff
@@ -457,6 +457,28 @@ class EstimatedPeriodConsumption < MeterAttributeTypes::AttributeBase
...
+      structure: {
+        per:  MeterAttributeTypes::Symbol.define(allowed_values: [:kwh]),
+        rate: MeterAttributeTypes::Float.define,
+        from: MeterAttributeTypes::TimeOfDay.define,
+        to:   MeterAttributeTypes::TimeOfDay.define
+      }
+    )
+  end
+
   class EconomicTariff < MeterAttributeTypes::AttributeBase
```
ldodds, line 482: This is breaking change for the existing EconomicTariff data.

The structure has moved from

```
 rates: { 
   rate: { per: ..., rate ...},
   daytime_rate:  { per: ..., rate ...},
   nighttime_rate:  { per: ..., rate ...}
 }
```

To the following:

```
 rates: { per: ..., rate ...},
 daytime_rate:  { per: ..., rate ...},
 nighttime_rate:  { per: ..., rate ...}
```

This means all of the existing tariffs are now invalid/broken. They fail to display in the application because they're not parsed properly by the updated code. Can you change the structure to match the existing data and align the new meter attribute to do the same thing, so the only difference is extra start/end date in the new attribute.

app/models/tariffs/meter_tariff.rb
```diff
@@ -50,6 +49,43 @@ def self.format_time_range(rate)
...
+  end
+
+  private
+
+  def find_tariff(date)
+    @tariffs.each do |date_range, tariff|
+      return tariff if date >= date_range.first && date <= date_range.last
+    end
+
+    nil # should fall over upstream
```
ldodds, line 77: Where is the upstream check for this? 

It looks like if this happens then both the `rate` and `weighted_cost` methods will end up generating an "undefined method  X for nil:NilClass" error? Would be better to explicitly rate an exception here with a useful message. Makes fault finding easier.

app/models/tariffs/meter_tariff_manager.rb
```diff
@@ -187,6 +187,16 @@ def pre_process_tariff_attributes(meter)
     check_tariffs
   end
 
+  def pre_process_economic_tariffs(meter)
+    if meter.attributes(:economic_tariff_change_over_time).nil? || meter.attributes(:economic_tariff_change_over_time).empty?
```
ldodds, line 191: Could use `.blank?` here. `nil` and an empty Array or Hash are always blank.

app/models/tariffs/meter_tariff_manager.rb
```diff
@@ -187,6 +187,16 @@ def pre_process_tariff_attributes(meter)
...
+      # pre October 2022 version of economic tariffs - same value for all time
+      EconomicTariff.new(meter, meter.attributes(:economic_tariff))
+    else
+      # post October 2022 version of economic tariffs - tariffs potentiall change with start, end_date over time
+      EconomicTariffChangeOverTime.new(meter, meter.attributes(:economic_tariff_change_over_time))
```
ldodds, line 196: The addition of a new attribute rather than extending the existing one (or including the existing `economic_tariff` attributes in this list) means that for any group or school to use this feature we'll need to have also created copies of the existing default economic tariffs but as `economic_tariff_change_over_time` attributes. So managing 8 rather than 4 defaults, plus the group or school specific ones. This add some admin overheads. 

Could you not include the `economic_tariff` attributes in this list but rely on the default start/end dates being applied?

ldodds (changes requested):

app/models/tariffs/meter_tariff.rb
PhilipTB: I am a bit reluctant to put in an upstream check as the code performance is critical - so the current state is deliberate, any check will slow the code down perhaps better to parse the meter attributes when they are instantiated to guarantee consistency - the  tariffs can only be edited by the admin user, in an ideal world the front end should stop incorrectly configured meter attributes but it doesn't, throwing an exception a long way downstream in the analytics for incorrectly added data from the admin user is less than ideal.

ldodds: Agree that there could be more upstream checks. But I'm unclear how raising an error if there's no tariff found, rather than relying on the very next line called to throw a NilClass error will impact performance here?

I'm suggesting replacing `nil` with a `raise`.

PhilipTB: Have added checking for tariff consistency on parsing/instantiation of tariffs

app/models/meter_attributes.rb
PhilipTB: Have made an attempt at fixing this. The analytics don't use these structures, just used for front end editing, so have guessed a little but should now be correct.

ldodds: The front-end presents its editing interface based on the structure provided in the analytics. So changes to structure can't be made without changes to the current database contents too. The analytics does rely on the structure, e.g. there's code in `meter_tariff.rb` that indexes in the hash using the `:rates` key.

E.g. 

```
  def differential?(_date)
    tariff[:rates].key?(:nighttime_rate)
  end
```

So this check could have been failing in the previous structure.

The attributes appear to be parsing correctly now though.

PR #351, opened 2022-11-02
Fix temperature range error for targets

iant:
New error occurring since we further reduced the window for schedule data:
https://rollbar.com/energysparks/EnergySparksTestEnvironment/items/2287/?item_page=0&item_count=100&#instances
This is from this code in [the temperatures class](https://github.com/Energy-Sparks/energy-sparks_analytics/blob/master/lib/dashboard/data_sources/utilities/temperatures.rb#L42). The code checks for a zero value in the previous line and logs if that happens, but then does a divide by zero.

The root cause though is [this line](https://github.com/Energy-Sparks/energy-sparks_analytics/blob/master/lib/dashboard/modelling/targeting%20and%20tracking/target_meter_temperature_compensated.rb#L68). It attempts to calculate an average temperature around a day (TimeOfYear) using 4 days either side.

There's an error when it attempts to calculate the average temperature for 2nd Feb as there isn't 4 days before this date available.

This PR:
- [x] sets a new constant for the days either side value (also used by the main application - see below)
- [x] raises an error when average temperature data is missing instead of raising for a divide by zero
- [x] Fixes a bug in `HalfHourlyData#missing_dates` so that it returns dates that are missing, not a collection of dates for which there is data

Note: there is a related pr/fix on the application side in https://github.com/Energy-Sparks/energy-sparks/pull/2160/files that looks for the newly defined `TARGET_TEMPERATURE_DAYS_EITHER_SIDE` constant (or defaults to 4) so we always pull in maximum meter date - 4 days.

### DIFF
--- a/lib/dashboard/data_sources/utilities/temperatures.rb
+++ b/lib/dashboard/data_sources/utilities/temperatures.rb
@@ -32,9 +32,7 @@ class Temperatures < HalfHourlyData
     while date >= start_date
       time_of_year_date_this_year = Date.new(date.year, time_of_year.month, time_of_year.day)
       if time_of_year_date_this_year.between?(start_date + days_either_side, end_date - days_either_side)
-        avg_temperatures += (-days_either_side..days_either_side).map do |offset|
-                                                                    average_temperature(time_of_year_date_this_year + offset)
-                                                                  end
+        avg_temperatures += (-days_either_side..days_either_side).map { |offset| average_temperature(time_of_year_date_this_year + offset) }
       end
       date = Date.new(date.year - 1, date.month, date.day)
     end
--- a/lib/dashboard/data_sources/utilities/temperatures.rb
+++ b/lib/dashboard/data_sources/utilities/temperatures.rb
@@ -1,6 +1,8 @@
 class Temperatures < HalfHourlyData
   include Logging
 
+  class MissingTemperatureDataError < StandardError; end
+
   FROSTPROTECTIONTEMPERATURE = 4.0
   def initialize(type)
     super(type)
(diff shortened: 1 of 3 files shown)


lib/dashboard/data_sources/utilities/temperatures.rb
```diff
@@ -31,15 +31,19 @@ def average_temperature_for_time_of_year(time_of_year:, days_either_side: 0)
...
 
-    logger.info "No data for time of year #{time_of_year}, have temperatures between #{start_date} and #{end_date}" if avg_temperatures.length == 0
-
-    avg_temperatures.sum / avg_temperatures.length
+    if avg_temperatures.length.zero?
+      logger.info "No data for time of year #{time_of_year}, have temperatures between #{start_date} and #{end_date}"
+      nil
+    else
+      avg_temperatures.sum / avg_temperatures.length
+    end
```
iant, line 46: hi @PhilipTB just to confirm, does this need a small refactor to return a better error if there is insufficient data (here i return nil instead of div by zero) or to use just the last date of data available?

(cc @ldodds)

iant: (@PhilipTB also, as per your other comment, I've updated the main application code to load 4 days extra data in this por https://github.com/Energy-Sparks/energy-sparks/pull/2160)

ldodds: Looking in the calling code, a response of `nil` will result in a `nil can't be coerced into Integer` error from this line, which is where the result is used:

https://github.com/Energy-Sparks/energy-sparks_analytics/blob/f4a7c5b604c14cec93788aafd4918754aca9f67e/lib/dashboard/modelling/targeting%20and%20tracking/target_meter_temperature_compensated.rb#L54

Returning 0 will mean there is no error, but will result in the target using a temperature of 15.5. Which is on the low side for the national average I think.

I'd suggest we either do that, or maybe raise an exception.

My original suggestion was to use what data we had if there wasn't 4 days before hand. E.g. reduce the window size used for the average. If there's no earlier data then just taking the temperature of the same day last year is arguably better than a complete failure in this case.

iant: thanks @ldodds  - I've pushed an update so it raise's an exception (was suggested by @PhilipTB too).

ldodds (approved):

PR #356, opened 2022-11-04
Migrate analytics constants for prices to the application

iant:
The BenchmarkMetrics class in the analytics includes some hard-coded prices:

```
  ELECTRICITY_PRICE = 0.15
  SOLAR_EXPORT_PRICE = 0.05
  GAS_PRICE = 0.03
  OIL_PRICE = 0.05
```

We want to move away from hard-coded prices and instead allow these to be managed by an EnergySparks admin via the application.  See https://trello.com/c/pcqvWgdW/2550-migrate-analytics-constants-for-prices-to-the-application

Adding a new admin interface and supporting database model to manage a list of prices has been started in https://github.com/Energy-Sparks/energy-sparks/pull/2170

This pr is a spike to see how refactoring the analytics code can be changed so that the current constants are replaced with a price lookup.  The analytics code needs to also fallback to using hard-coded values if not called by the Rails application.

The above list of prices may also be extended over time as there's other parts of the analytics, e.g. `EnergyEquivalences` which also have hard-coded prices for school meals, teaching assistant hourly rates, etc.


lib/dashboard/alerts/common/benchmark_metrics.rb
```diff
@@ -20,8 +20,27 @@ module BenchmarkMetrics
   ANNUAL_AVERAGE_DEGREE_DAYS = 2000.0
   AVERAGE_GAS_PROPORTION_OF_HEATING = 0.6
 
-  BENCHMARK_ENERGY_COST_PER_PUPIL = BENCHMARK_GAS_USAGE_PER_PUPIL * GAS_PRICE +
-                                    BENCHMARK_ELECTRICITY_USAGE_PER_PUPIL * ELECTRICITY_PRICE
+  # BENCHMARK_ENERGY_COST_PER_PUPIL = BENCHMARK_GAS_USAGE_PER_PUPIL * BenchmarkMetrics.pricing.gas_price +
```
iant, line 23: @PhilipTB `BENCHMARK_ENERGY_COST_PER_PUPIL` doesn't seem to be used anywhere - can this be removed now?

PhilipTB: yes, if there is no reference to it.

iant: :+1: thanks @PhilipTB

lib/dashboard/alerts/common/benchmark_metrics.rb
```diff
@@ -20,8 +20,27 @@ module BenchmarkMetrics
...
+  #                                   BENCHMARK_ELECTRICITY_USAGE_PER_PUPIL * ELECTRICITY_PRICE
+
+
+  def self.pricing
+    if Object.const_defined?('Rails')
+      SiteSettings.current_prices
+    else
+      OpenStruct.new(
```
iant, line 27: think probably better here to use a hash or Struct.  Same for `SiteSettings#current_prices`.

iant:
See also https://github.com/Energy-Sparks/energy-sparks/pull/2170 - which would need to be be merged first.

ldodds (approved):
I think this looks OK. We'll need to test it to make sure there's no performance issues, but not sure how else we could approach this.

iant:
Putting a blocked label on this one for now (and re-marking as draft) while @PhilipTB takes a look at what might needs doing in the analytics library to use alternative tariffs.

PR #360, opened 2022-11-17
Sort out poor labelling of meter breakdown charts when there is solar…

iant:
… (attempt 2)


iant:
closing this in favour of new pr

PR #362, opened 2022-11-22
Fix poor labelling of meter breakdown charts when there is solar

iant:
Note: this is with meter names updated locally 
irb(main):023:0> s = School.find_by(name: "***********")
irb(main):024:0> s.meters.map { |m| [m.id, m.mpan_mprn, m.name] }
=> [[338, 72199999800132, "School hall"], [339, 62199999800132, "Main classroom"], [330, 2199999800132, "Sports hall"], [375, 92199999800132, "Staffroom"]]

Screenshot 2022-11-22 at 12 32 04


ldodds:
@iant can we revive this one and get it merged? Suggest asking Claudia for any feedback on the meter naming if needed. We could always add a chart footnote now to indicate that meters include solar consumption as an alternative.

ldodds (approved):

PR #364, opened 2022-12-07
Check whether the date exists before trying to set start and end date

ldodds:
Admins can set the `readings_start_date` and `readings_end_date` attributes for a meter. The analytics then ignores data outside of that range when processing.

This PR adds a check for whether the date provided actually exists in the date. Currently if the admin chooses an incorrect date, or there is a gap in the range for this meter, then we just get a `NoMethodError`.

Instead, add an explicit check for the start/end date and throw an exception with a more meaningful message.

### DIFF
--- a/lib/dashboard/aggregation/validate_amr_data.rb
+++ b/lib/dashboard/aggregation/validate_amr_data.rb
@@ -221,7 +221,10 @@ class ValidateAMRData
   end
 
   def check_date_exists(date, attribute)
-    raise EnergySparksMeterSpecification.new("Unable to apply #{attribute} correction for meter #{@meter_id}. There is no meter data for #{date}. Check the date and/or for missing readings") unless @amr_data.date_exists?(date)
+    return if @amr_data.date_exists?(date)
+
+    error_message = "Unable to apply #{attribute} correction for meter #{@meter_id}. There is no meter data for #{date}. Check the date and/or for missing readings"
+    raise EnergySparksMeterSpecification.new(error_message)
   end
 
   def extend_start_date(date)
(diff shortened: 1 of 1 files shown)


lib/dashboard/aggregation/validate_amr_data.rb
```diff
@@ -212,6 +204,26 @@ def apply_one_meter_correction(rule)
...
+  def apply_readings_end_date(fix_end_date)
+    check_date_exists(fix_end_date, :readings_end_date)
+    logger.debug "Fixing end date to #{fix_end_date}"
+    substitute_data_x48 = @amr_data.one_days_data_x48(fix_end_date)
+    @amr_data.add(fix_end_date, OneDayAMRReading.new(meter_id, fix_end_date, 'FIXE', nil, DateTime.now, substitute_data_x48))
+    @amr_data.set_end_date(fix_end_date)
+  end
+
+  def check_date_exists(date, attribute)
+    raise EnergySparksMeterSpecification.new("Unable to apply #{attribute} correction for meter #{@meter_id}. There is no meter data for #{date}. Check the date and/or for missing readings") unless @amr_data.date_exists?(date)
```
iant, line 224: as this line's quite long perhaps we could have an early return here?  Something like:
```
def check_date_exists(date, attribute)
  return if @amr_data.date_exists?(date)

  error_message = "Unable to apply #{attribute} correction for meter #{@meter_id}. There is no meter data for #{date}. Check the date and/or for missing readings"
  raise EnergySparksMeterSpecification.new(error_message) 
end
```

iant (approved):
:+1: looks good to me @ldodds 1 comment/suggestion

PR #369, opened 2022-12-15
New Advice Page Baseload services

ldodds:
This is an initial prototype to illustrate how we will be developing updated service interfaces as part of the advice page development. It is not ready for merging as it needs to be aligned with the most recent cost-related changes to the analytics.

Currently the application uses the classes that are in the `AdviceBase` hierarchy (e.g. `AdviceBaseload`) to request that the analytics does the necessary calculations and then assembles the content to be displayed to the user. Running the calculations involves:

- running charts
- running alerts and extracting specific variables
- calculating additional values and summaries

As part of the translation work all of the presentation elements, e.g. deciding what text, charts and other content to appear on the appears will be done in the application.

Currently its not easy to extract all the necessary data to support that from the `AdviceBase` framework. So the intention is to create new service interfaces that expose the raw data to the application.

This PR provides three new services that covers the main data exposed on the current Baseload advice page:

- `Baseload::BaseloadCalculationService` - calculating baseload for a specific meter and associated usage data (`kwh`, `£`, co2`)
- `Baseload::BaseloadBenchmarksService` - for a given school, calculate the benchmark values used to indicate how a school can reduce its energy usage and costs
- `Baseload::BaseloadMeterBreakdownService` - for a given school, calculate the average baseload and cost associated with each electricity meter

What's not covered are services for doing seasonal and intraday variations, but these would follow the same pattern.

The services were developed by reviewing the existing `AdviceBaseload` class and checking what variables/calculations it relies on. See `AdviceBaseload.content`, `adult_dashboard_configuration.rb`, `AlertElectricityBaseloadVersusBenchmark`, `AlertBaseloadBase`, etc.

The services either return simple values, e.g. a `kw` value or custom objects (`CombinedUsageMetric`, `MeterBaseloadBreakdown`). The intention is to move towards returning objects, which may expose additional helper methods, rather than hashed values. This provides a more stable, testable interface for the application. These also provide the basic building blocks required to create the new pages. 

Internally the services all delegate to existing code in the analytics framework, e.g. the `ElectricityBaseloadAnalysis` classes. This allows the internals of the analytics framework to evolve separately to the interface exposed to the application.

Where necessary additional existing code from the advice or alerts framework can be refactored into additional supporting classes (see, e.g. `BlendedRateCalculator` as an example). This will allows us to move towards composition vs inheritance for shared behaviour. Again, this improves testability and simplifies the code.

New services should have rspec tests that confirm the basic behaviour. This PR initially uses an anonymised set of school data as a text fixture (a serialised `MeterCollection` taken from the application). Overtime we should be able to move towards using factories to assemble necessary data.


ldodds:
The PR has been updated to:

- rebase against master to pick up latest changes to alerts and advice pages
- updated all the services to ensure they're using the latest methods for baseload analysis and are calculating `£` costs consistent with the alerts. (the alerts have new `£current` vars, these aren't exposed via these services as they don't seem to be used by the `AdviceBaseload` page, we may need to revisit that)
- added a custom matcher to round returned values to 2 dp, as there were small, largely irrelevant, changes in small values at 10-12 decimal places

Still to do are:

- [ ] add service for seasonal calculations
- [ ] add service for intraday calculations
- [ ] any final tidy up of the methods

We can then merge this to support development of new baseload page.

jhigman (approved):

PR #373, opened 2022-12-23
Bump activesupport version

iant:
Bump activesupport gem version so we can investigate upgrading rails on the main application to 6.1.7


iant:
closing as looks like upgrading to rails 6.1.7 may be tricky https://trello.com/c/G9rTZJDq/2772-upgrade-rails-to-617

PR #374, opened 2023-01-05
Fix bug in one_years_data?

ldodds:
The `ElectricityBaseloadAnalysis.one_years_data?` method has a bug. Currently it only checks to see if there's a years worth of data for a meter, rather than a years worth of data preceding the `asof_date` which is provided as a parameter.

This PR fixes the method so it behaves as expected. Unlikely to be a problem in production, encountered the issue whilst running some tests with older dates.


ldodds:
@PhilipTB does this look OK to you?

PhilipTB (approved):

PR #387, opened 2023-01-13
Organise weekly baseload analysis

ldodds:
Adds a method to the ElectricityBaseloadAnalysis object for returning average baseload in last week.

Updated the new services and the existing alert to call this method, to ensure the logic remains consistent.


iant (commented):
looks good @ldodds just double checking the changes to script/standard/test_alerts.rb need to be in the pr?

iant (approved):

PR #390, opened 2023-01-17
Fix for thermostatic charts

ldodds:
A [recent change](https://github.com/Energy-Sparks/energy-sparks_analytics/commit/65779f7f97ef19adc15849f56ef865d75dbf0597#diff-07bfcde2ea5fa51db9042ab04b678fe44c3c87daf3031bbdb10c576c8db9bd31R409) broke the `predicted_amr_data_one_day` method in SeriesDataManager.

This didn't show up in the analytics test suite because that method only seems to be called from `AggregatorTrendlines` which is used to add trend lines. Different code is called for rails vs the excel output. So the analytics test suite didn't fail, but the application charts are broken.

This PR fixes the method call.


iant (approved):
@ldodds looks good to me

PR #393, opened 2023-01-18
Update rubocop yml

iant:


ldodds (approved):

iant:
@ldodds as discussed I've updated the rubocop config to exclude all but services and related files

ldodds:
We seem to be having to regularly add `rubocop:disable Naming/AsciiIdentifiers` to classes to allow for variables/methods that include the `£` sign. This seems so common across our code that I'm thinking we should just relax the requirements in rubocop as a whole in both projects.

I personally prefer using £ as less risk of a typo around gbp, gpb, etc.

iant:
> We seem to be having to regularly add `rubocop:disable Naming/AsciiIdentifiers` to classes to allow for variables/methods that include the `£` sign. This seems so common across our code that I'm thinking we should just relax the requirements in rubocop as a whole in both projects.
> 
> I personally prefer using £ as less risk of a typo around gbp, gpb, etc.

yep sounds good to me.  
```
Naming/AsciiIdentifiers:
  Enabled: false
```
has actually already been added to the `rubocop.yml` here so we should be able to remove those `rubocop:disable` comments now (if not auto removed when we run rubocop)

PR #395, opened 2023-01-20
Run rubocop on app services directory

iant:


ldodds (approved):

iant:
hi @ldodds I've added a couple of `rubocop:disable Lint/MissingSuper` comments to 
`app/services/baseload/intraweek_baseload_service.rb` and `app/services/baseload/seasonal_baseload_service.rb`, as per the overcommit rubocop warning.   I'll get this merged once passing but would you be able to double check these classes when you get a mo please.

PR #398, opened 2023-01-20
New services for thermostatic analysis

iant:


app/services/heating/heating_thermostatic_analysis_service.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module Heating
+  class HeatingThermostaticAnalysisService
+    attr_reader :aggregated_heat_meters
```
ldodds, line 5: probably don't need this?

app/services/heating/heating_thermostatic_analysis_service.rb
```diff
@@ -0,0 +1,100 @@
+# frozen_string_literal: true
+
+module Heating
+  class HeatingThermostaticAnalysisService
```
ldodds, line 4: Could this not extend the Heating::BaseService?

Looks like there's some shared behaviour, e.g. creating a model (but need to check whether the way that the `AdviceGasThermostaticControl` and now this class aligns with the model created in the code I wrote.

The `enough_data?` check will also allow the application to test whether we can actually create and run a model.

ldodds (changes requested):

ldodds:
Would be too to finish this up @iant. Can you review my comment about extending the base service I added, will help us avoid code duplication around model creation.

Overriding the `enough_data?` check (if necessary) will be part of adding the new Analysable mixin.

iant:
@ldodds I think this is ready for another review.  heating model seems to differ to the one in the base service so have overridden but we can still use the enough_data and validate_meter_collection methods.

app/services/heating/heating_thermostatic_analysis_service.rb
```diff
@@ -0,0 +1,105 @@
...
+      average_outside_temperature_high_centigrade: 12.0,
+      average_outside_temperature_low_centigrade: 4.0
+    )
+      validate_meter_collection(meter_collection)
+      @meter_collection = meter_collection
+      @average_outside_temperature_high_centigrade = average_outside_temperature_high_centigrade
+      @average_outside_temperature_low_centigrade = average_outside_temperature_low_centigrade
+    end
+
+    def data_available_from
```
ldodds, line 18: This isn't correct. The AnalysableMixin data_available_from method should be the date from which the analysis will be available, i.e. when there is sufficient data available, not the date from which we have data.

I've implemented some generic code for this check here:

https://github.com/Energy-Sparks/energy-sparks_analytics/pull/404/files#diff-5ea50f622f59bb6ebe5351661d00cc9f981910fcb0b247b328ff18d717e41ff4R66

And use it here:

https://github.com/Energy-Sparks/energy-sparks_analytics/pull/404/files#diff-9d1273e9c4b66026620637517f4775672860bd2accc3ac42bd8651baefd1a14dR28

ldodds: Maybe go ahead and merge this, when review these two methods across all services once we've merged my PR.

ldodds (approved):

PR #399, opened 2023-01-24
Implement new services for hot water analysis

iant:
Screenshot 2023-01-25 at 14 20 01
Screenshot 2023-01-25 at 14 20 12

### DIFF
--- a/app/services/hot_water/gas_hot_water_service.rb
+++ b/app/services/hot_water/gas_hot_water_service.rb
@@ -1,0 +1,11 @@
+module HotWater
+  class GasHotWaterService
+    def initialize(aggregated_heat_meters:)
+      @aggregated_heat_meters = aggregated_heat_meters
+    end
+
+    def create_model
+      {}
+    end
+  end
+end
--- a/app/services/hot_water/gas_hot_water_service.rb
+++ b/app/services/hot_water/gas_hot_water_service.rb
@@ -1,11 +1,40 @@
 module HotWater
   class GasHotWaterService
-    def initialize(aggregated_heat_meters:)
-      @aggregated_heat_meters = aggregated_heat_meters
+    def initialize(meter_collection:)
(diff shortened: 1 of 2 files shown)


app/services/hot_water/gas_hot_water_service.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module HotWater
+  class GasHotWaterService
+    def initialize(meter_collection:)
+      @meter_collection = meter_collection
+    end
+
+    def create_model
```
ldodds, line 9: I think I would prefer to see different methods for accessing the investment choices and the efficiency breakdowns rather than lumping all the calculations together. We may not always want to calculate both together.

app/services/hot_water/gas_hot_water_service.rb
```diff
@@ -0,0 +1,53 @@
...
+      investment_analysis_annual.each_with_object(OpenStruct.new) do |(type, values), investment_choices|
+        investment_choices[type] = OpenStruct.new(values)
+      end
+    end
+
+    def hotwater_model
+      investment_analysis.hotwater_model
+    end
+
+    def efficiency_breakdowns
```
ldodds, line 37: This is producing a very similar day type breakdown as the `UsageBreakdownService`. In that class we're returning an object (`UsageCategoryBreakdown`). It looks like we're returning calculating different variables here, and also across daily/annual. But any reason not to align what we're doing across these services?

IMO, one of the main benefits of returning an object rather than just an OpenStruct is that its much clearer about what is being returned from some chunk of analysis. We can document the data structure and ensure that we're using the right variables. At the moment I had to read the rspec in order to get an idea about what was returned here.

Using objects, rather than open structures reduces likelihood that when we stub things for testing in the application we end up with mismatched expectations around what is returned.

ldodds (changes requested):

PR #401, opened 2023-01-27
Add "Analysable" mixin to analytics services

iant:


ldodds (approved):
Looks good. Where you've added the comments, you could move them to be proper method/class comments, so if we decide to start running rdoc, we'll get some useful output.

PR #402, opened 2023-01-27
New service for recent progress

iant:


app/services/baseload/baseload_period_comparison_service.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module Baseload
```
ldodds, line 3: I think this is the wrong module and class name? 

The recent changes advice page isn't doing any comparisons for baseload, it's looking at overall energy consumption for gas and electricity.

app/services/baseload/baseload_period_comparison_service.rb
```diff
@@ -0,0 +1,66 @@
...
+  class BaseloadPeriodComparisonService < BaseService
+    def initialize(meter_collection:, fuel_type:, date: Date.today)
+      @meter_collection = meter_collection
+      @fuel_type = fuel_type
+      @date = date
+    end
+
+    def create_model
+      OpenStruct.new(
+        current_period_start_date: current_period_start_date,
```
ldodds, line 13: These should really be a pair of CombinedUsageMetric, one for the current period, one for the previous. Along with the period start and end date. 

It might be useful to have a Period Comparison class that covers that, we could use that in other places to I think.

iant:
@ldodds this is ready for a review.  docs
https://docs.google.com/document/d/1y8vGPUutYCkM-uVGkB6BFQhPPMjIq75_8qpxg5OOm_M/edit
https://docs.google.com/document/d/1gYLAVDLOit94k6kZLSJOqZ3D97WJZIoY02SXFATETFE/edit#heading=h.bxsn344l7y99
ratings and an equivalences are missing (difference_co2_smartphone_kwh), which i assume can be done on the application side?

app/services/usage/combined_usage_metric_comparison_service.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Usage
+  class CombinedUsageMetricComparisonService
```
ldodds, line 4: This isn't a service, its a plain old ruby object?

Can you rename as `CombinedUsageMetricComparison` and move to the `classes` directory?

app/services/usage/recent_usage_comparison_service.rb
```diff
@@ -0,0 +1,87 @@
...
+
+module Usage
+  class RecentUsageComparisonService
+    def initialize(meter_collection:, fuel_type:, date: Date.today)
+      @meter_collection = meter_collection
+      @fuel_type = fuel_type
+      @date = date
+    end
+
+    def create_model
```
ldodds, line 11: maybe `recent_usage` as more descriptive, we're not creating a model here.

ldodds (changes requested):

iant:
@ldodds class and method name updated. think this one's ready for another review....

ldodds (approved):

PR #403, opened 2023-01-30
Fix issues with annual electricity advice page

ldodds:
The Annual Electricity advice page was not showing for a school. Debugging showed that the `AlertElectricityAnnualVersusBenchmark` alert was recently changed to add calculations that need 2 years of data, not just 1. These variables are only used on the school comparison pages. 

So the alert fails for schools that have >1 year but less than 2 years of data. 

This then causes the advice page to also fail.

This PR introduces an initial fix that checks whether the extra year of data is available. If not, then it skips those variables. The school comparison page already ignores schools where these vars are nil. The variables would be better calculated in a different alert which already checks for 2 years worth of data.

Whilst debugging this I noticed that `AdviceBase` can throw an exception when generating output. It iterates through its list of charts and attempts to format their output after running them. But it wasn't checking for `nil` entries in the list of charts. A `nil` entry is added if a chart fails to run if there's not enough data (which happens here as one chart needs at least 2 years data) or some other error occurs. This immediately leads to a `StandardError`.

As the failed charts are already logged by the test framework, I've just added a `.compact` call to avoid trying to produce content for failed charts.


iant (approved):
looks good to me @ldodds

PR #404, opened 2023-01-31
Annual use services

ldodds:
Adds new services:

- `Usage::AnnualUsageCalculationService` to calculate annual usage, and also comparison with previous year
- `Usage::AnnualUsageBenchmarksService` to calculate benchmarks and estimated savings vs benchmark
- `Usage::AnnualUsageMeterBreakdownService` to calculate breakdown by meter
- Renamed the `UsageBreakdownService` as `AnnualUsageBreakdownService` and moved to `Usage` module
- Similar changes to the breakdown object returned by above
- Move blended rate calculator to separate module
- Add `Schools::SchoolSizeCalculator` to wrap calls to calculate number of pupils and floor area
- Adds `Meters::MeterDataRangeChecker` to provide methods checking how much data we have for a meter, when we'll have enough data to run calculations and whether data is lagging

The Annual Usage Meter Breakdown Service diverges from the other services in that it uses custom logic to work out the start/end date for analysis. This is because the current implementation is intended to align its dates with a specific chart, rather than just calculate the last 12 months of data. It also applies some different logic for deciding whether to include a meter in the breakdown. 

For the moment I've aligned this code with the current implementation. I've commented this extensively in the class. We might want to revisit that later, or update the service so that it can calculate either a 12 month breakdown or a summary of the existing chart.


app/services/meters/meter_data_range_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+    # Specifying a nil asof_date means the checks will be done against the
+    # latest date for which we have data for this meter.
+    #
+    # If the date to be checked is later than the latest data, then checks will
+    # be done against that date
+    #
+    # @param [Dashboard::Meter] analytics_meter the meter to be checked
+    # @param [Date] asof_date the date to use as the basis for calculations
+    # @param [Integer] days_in_year we sometimes treat a year as 364 days, allow override
+    def initialize(analytics_meter, asof_date=Date.today, days_in_year: 365)
```
iant, line 17: not something for fixing in this pr but maybe for consistency we should try and use keyword arguments across all classes?  perhaps something to add to our review at the end?

app/services/usage/annual_usage_benchmarks_service.rb
```diff
@@ -0,0 +1,168 @@
...
+    #
+    # @raise [EnergySparksUnexpectedStateException] if the schools doesnt have electricity meters
+    def initialize(meter_collection, fuel_type, asof_date=Date.today)
+      validate_meter_collection(meter_collection, fuel_type)
+      @meter_collection = meter_collection
+      @fuel_type = fuel_type
+      @asof_date = asof_date
+    end
+
+
```
iant, line 21: looks like an extra line here?

app/services/usage/annual_usage_calculation_service.rb
```diff
@@ -0,0 +1,110 @@
+module Usage
+  class AnnualUsageCalculationService
+    DAYSINYEAR = 363
```
iant, line 3: as DAYSINYEAR is used in more than one place maybe it could go in the AnalysableMixin?

app/services/usage/annual_usage_calculation_service.rb
```diff
@@ -0,0 +1,110 @@
...
+    #
+    # If there isn't sufficient data (>2 years) then the method will return nil
+    #
+    # @return [CombinedUsageMetric] the difference between this year and last year
+    def annual_usage_change_since_last_year
+      return nil unless has_full_previous_years_worth_of_data?
+      this_year = annual_usage(period: :this_year)
+      last_year = annual_usage(period: :last_year)
+      kwh = this_year.kwh - last_year.kwh
+      return CombinedUsageMetric.new(
```
iant, line 64: return not needed here

app/services/usage/annual_usage_benchmarks_service.rb
```diff
@@ -0,0 +1,168 @@
...
+    #Do we have enough data to run the calculations?
+    def enough_data?
+      meter_data_checker.one_years_data?
+    end
+
+    #If we don't have enough data, then when will it be available?
+    def data_available_from
+      meter_data_checker.date_when_enough_data_available(365)
```
iant, line 29: could 365 be a constant?

app/services/meters/meter_data_range_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def at_least_x_days_data?(days)
+      days_of_data >= days
+    end
+
+    #How many days of data to be have between the date we're
+    #checking and the start data of the meter
+    def days_of_data
+      @asof_date - @meter.amr_data.start_date + 1
```
iant, line 38: just double checking if brackets might be needed here e.g.
```
irb(main):013:0> (Date.new(2020,12,10) - (Date.new(2020,12,8) + 1)).to_i
=> 1
irb(main):014:0> (Date.new(2020,12,10) - Date.new(2020,12,8) + 1).to_i
=> 3
```

app/services/usage/annual_usage_benchmarks_service.rb
```diff
@@ -0,0 +1,168 @@
...
+
+    def degree_day_adjustment
+      BenchmarkMetrics.normalise_degree_days(@meter_collection.temperatures, @meter_collection.holidays, :gas, @asof_date)
+    end
+
+    def percent_change(old_value, new_value)
+      return nil if old_value.nil? || new_value.nil?
+      return 0.0 if !old_value.nan? && old_value == new_value # both 0.0 case
+      (new_value - old_value) / old_value
+    end
```
iant, line 126: I've used this same method in my pr https://github.com/Energy-Sparks/energy-sparks_analytics/pull/402/files#diff-ccc5a444b9ef6ec32e02b3f6268f2e88b277e975719831c2f8bd9e9c0810b79dR25 - perhaps this is something we can move to the analysable mixin (or somewhere) later (and write a unit test for it too)

iant (approved):
looks good to me @ldodds - a few comments but nothing major so it's good to go i think

app/services/usage/annual_usage_calculation_service.rb
ldodds: I'm not sure it fits there and I think I'd like to just tidy all these up at some point.

app/services/usage/annual_usage_benchmarks_service.rb
ldodds: Yes, I think we can move this somewhere. But would prefer a general helper class I think rather than the Analysable mixin. That has a specific purpose.

PR #415, opened 2023-02-13
Add one year only version of benchmark charts, defaulting to kwh

ldodds:
Adds versions of the benchmark charts limited to just one year of data and defaulting to kwh rather than £.


iant (approved):
looks good to me @ldodds

PR #416, opened 2023-02-13
Fix method name in usage benchmark service

ldodds:
Method name referred to "electricity" but service is generic across fuel types. Just renames the method.

Will require application side change to call the new method once merged and integrated.


iant (approved):
:+1:

PR #419, opened 2023-02-15
Refactor seasonal analysis service

ldodds:
* Rename the HeatingBoilerSeasonalControlAnalysisService to SeasonalControlAnalysis
* Add base class
* Implement the `enough_data?` check to customise for this class
* Rename main method and rework structure to separate out estimated savings from other data
* Added some comments, to help clarify the calculation (at least for me!)
* Refactored the seasonal analysis further to allow extraction of additional data for calculating percent of annual heating
* Added percent of annual heating value to result

One thing I noted here is that there are some minor differences between how the existing boiler control advice page (`AdviceGasBoilerSeasonalControl`) and the `AlertSeasonalHeatingSchoolDays` alerts calculate heating models:

The existing advice page creates a heating model and directly extracts some data from it to calculate heating on days, etc. While it runs the `AlertSeasonalHeatingSchoolDays` alert it doesn't use either of the promoted variables in populating the page contents.

When creating a heating model:

Advice page:

- uses up to 12 months of data prior to the aggregate gas meter end date (`aggregate_meter.amr_data.end_date - 365`) when creating the `SchoolDatePeriod` for the heating model
- does not specify an `end_date` when calling `heating_model.heating_on_seasonal_analysis`, so this defaults to the aggregate gas meter end date 

Alert:

- uses up to 12 months of data prior to the `asof_date` (`asof_date - 364`) when creating the `SchoolDatePeriod` for the heating model. This is 1 day difference
- specifies an `end_date` when calling `heating_model.heating_on_seasonal_analysis`, which is `asof_date`

In practice the end dates are likely to align in the application. Although when running test scripts the dates may vary depending on what is is `test_alerts`.

But the `SchoolDatePeriod` are slightly different as the periods have a one day difference in data. This might not matter much in practice, especially when data is rounded for display.


iant (approved):
looks good to me @ldodds

PR #420, opened 2023-02-16
Analysable mixin for baseload and heating services

ldodds:
Ensure the AnalysableMixin interface is properly implemented across the baseload and heating services (other modules tbd)

* Delegate to MeterDateRangeChecker wherever possible, for consistency
* Add base class methods for constructing checker and providing default checks
* Add specific checks for the individual services
* Add rspecs to check the behaviour


iant (approved):
:+1: looks good to me @ldodds

PR #423, opened 2023-02-20
Return the right chart for school week comparison gas alert

ldodds:
Gas alert was returning the electricity chart.


iant (approved):
looks good to me @ldodds 

(not something for this pr but, to stick with convention, we should probably be underscoring
`lib/dashboard/alerts/time period comparison/` & `lib/dashboard/alerts/storage heaters/` directories)

PR #425, opened 2023-02-21
Add analysable mixin to existing service classes (Usage)

iant:


iant:
hi @ldodds I think this one's ready for review.  I'm not sure if we can add analysable to `RecentUsagePeriodCalculationService` as we're unsure of the date range(period) when the class is initialised but if you have any suggestions that would be great.

ldodds (approved):

ldodds (approved):

PR #426, opened 2023-02-23
Clamp negative hotwater percentages to 0%

iant:


iant:
marking this blocked pending decision to simply hide the page if negative pricing is shown

PR #427, opened 2023-02-23
Add Analysable to hot water service

iant:


iant:
note this still needs testing with a school with not enough gas data.  will confirm before merging.

iant:
@ldodds  this has been tested with schools with and without enough data.  We need to rescue from the error raised in `AnalyseHeatingAndHotWater::HotwaterModel#analyse_hotwater_around_summer_holidays` so i think this is ok to review now.

ldodds (approved):

PR #428, opened 2023-02-23
Add missing benchmark co2 values

ldodds:
The benchmark and exemplar values for the benchmarking charts (e.g. `:benchmark`) have recently been changed to extract values from relevant alerts.

However the alerts were not producing the necessary values for co2, which meant that the charts were not showing the expected comparisons.

This PR updates the alerts to add in the missing values. I've confirmed that the charts now work correctly.

A couple of co2 values were being calculated & these changes use the same method, so hopefully consistent.

@PhilipTB I'd appreciate it if you could take a look.

### DIFF
--- a/lib/dashboard/alerts/common/alert_analysis_base.rb
+++ b/lib/dashboard/alerts/common/alert_analysis_base.rb
@@ -496,7 +496,6 @@ class AlertAnalysisBase < ContentBase
     asof_date = [asof_date, alert.aggregate_meter.amr_data.end_date].min
 
     alert.analyse(asof_date)
-
     alert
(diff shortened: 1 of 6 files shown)


iant (approved):
looks good to me :+1:

ldodds:
@PhilipTB does this look OK to you?

PhilipTB:
> 

Yes looks ok but perhaps need to review the resulting numbers?

ldodds:
@PhilipTB OK, I'll merge and release to test tomorrow morning. The figures will show up on the charts, so will share a link.

PR #430, opened 2023-03-01
Add I18n support for Benchmark content - Energy Cost Benchmarks

iant:
This pr covers extracting text for translation for classes in the "Energy cost benchmarks" group
```
      {
        name:       'Energy Cost Benchmarks',
        benchmarks: %i[
          annual_energy_costs_per_pupil
          annual_energy_costs
          annual_energy_costs_per_floor_area
        ]
      },
```


spec/lib/dashboard/benchmarking/benchmark_content_base_spec.rb
```diff
@@ -0,0 +1,280 @@
...
+                            'Last year gas kWh/pupil' => [nil, 567.4328947285189, 596.2005734714177],
+                            'Last year storage heater  kWh/pupil' => [nil, nil, nil] },
+                  chart1_type: :bar,
+                  chart1_subtype: :stacked,
+                  y_axis_label: 'kWh',
+                  config_name: 'annual_energy_costs_per_pupil' } },
+     { type: :html, content: ' <h3>Chart interpretation</h3> ' },
+     { type: :html, content: ' <h3>Table Introduction</h3> ' },
+     { type: :table_html,
+       content: "\n" \
```
urbanwide, line 50: I am sure you've looked at options already but I wonder if this can be done in a more readable way. I'm thinking a heredoc - something like this (although I appreciate this will be a pain to change now, so feel free to ignore me!):
```
content: <<-STRING,
  <table class="table table-striped table-sm">

    <thead>
    .. etc ..
STRING
```

iant: yep good point @urbanwide that test isn't really adding anything useful so I think I'll just remove it entirely in one of the follow up pr's (another 5 to go).

spec/lib/dashboard/benchmarking/benchmark_content_base_spec.rb
```diff
@@ -0,0 +1,280 @@
...
+                           'Last year storage heater  kWh/pupil',
+                           'Last year energy kWh/pupil',
+                           'Last year energy £/pupil',
+                           'Last year energy kgCO2/pupil',
+                           'Type'],
+                  rows: [[{ formatted: 'Acme Secondary School 3',
+                            raw: 'Acme Secondary School 3',
+                            urn: 123_678,
+                            drilldown_content_class: AdviceBenchmark },
+                          { formatted: '362', raw: 361.7098785425103 },
```
urbanwide, line 234: I might be misunderstanding Analytics here but I was wondering what would happen if the underlying calculations change - are some of these raw figures subject to change too? Just thinking it might be a bit of a mare to change them all if this does happen

spec/lib/dashboard/benchmarking/benchmark_content_base_spec.rb
urbanwide: OK 😄  I feel your pain. I'll approve then and you can remove the test later?

urbanwide (approved):
Looks good @iant 👍

spec/lib/dashboard/benchmarking/benchmark_content_base_spec.rb
iant: :+1: yep will do. thanks @urbanwide !

PR #431, opened 2023-03-01
Add I18n support for Benchmark content - Change in Energy Use

iant:
This pr 
- [x] improves the test coverage of the 3 Energy Cost Benchmarks
- [x] adds a new html matcher so we can write tests that translated benchmark html content matches expected text
- [x] adds translation support for content_title etc


config/locales/benchmarking/content_base.yml
```diff
@@ -30,6 +30,267 @@ en:
...
+          unadjusted: Unadjusted
+          unadjusted_kwh: Unadjusted change (kWh)
+          urn: URN
+          variation_in_baseload_between_days_of_week: Variation in baseload between days of week
+          w_floor_area: w/floor area
+          weekend: Weekend
+          winter_baseload_kw: Winter baseload kW
+          year_before_joined: Year before joined
       content:
         annual_energy_costs:
```
ldodds, line 295: I'm wondering whether for the content for individual benchmarks, they might be better in specific files? Might make them easier to manage? What do you think?

lib/dashboard/benchmarking/benchmark_configuration.rb
```diff
@@ -383,6 +383,7 @@ def self.blended_baseload_rate_col(variable)
       annual_energy_costs_per_pupil: {
         benchmark_class:  BenchmarkContentEnergyPerPupil,
         name:     'Annual energy use per pupil',
+        i18n_key: :annual_energy_costs_per_pupil,
```
ldodds, line 386: The benchmark classes have a `page_name` attribute which is the name of the benchmark in the configuration.

I think it would make sense to use that, rather than an additional key to look up the translations. In the changes here, the key is always the same as the name anyway.

ldodds (changes requested):

config/locales/benchmarking/content_base.yml
iant: :+1: yep, definitely.  I'll move to specific files in a follow up pr.

ldodds (approved):

PR #449, opened 2023-03-07
Additional cost services

ldodds:
Adds some additional services for displaying costs page:

* `AccountingCostsService` to return the annual costs associated with a meter, plus change from previous year
* `TariffInformationService` to summarise coverage of tariffs and available data
* `AgreedSupplyCapacityService` to calculate how close a school is to their ASC limit. This is still a bit of a work in progress as there's no production data that uses this
* Some minor reorganisation of files

Need to merge this to update the application with latest changes to cost page.

Further refactoring of the costs services to come in future PR.


app/services/costs/accounting_costs_service.rb
```diff
@@ -0,0 +1,108 @@
...
+      enough_data? ? nil : meter_data_checker.date_when_enough_data_available(365)
+    end
+
+    #return cost, up to one year
+    def annual_cost
+      start_date = [@meter.amr_data.end_date - DAYSINYEAR, @meter.amr_data.start_date].max
+      #FIXME end date?
+      cost = @meter.amr_data.kwh_date_range(start_date, @meter.amr_data.end_date, :accounting_cost)
+      days = @meter.amr_data.end_date - start_date + 1
+      return OpenStruct.new(
```
iant, line 26: return not needed

app/services/costs/agreed_supply_capacity_service.rb
```diff
@@ -0,0 +1,51 @@
...
+        cost_£ = 12.0 * asc_limit_kw * agreed_availability_charge[:rate]
+        saving_£ = (1.0 - kw / asc_limit_kw) * cost_£
+      end
+
+      AgreedSupplyCapacitySummary.new(
+        kw: kw,
+        agreed_limit_kw: accounting_tariff.tariff[:asc_limit_kw],
+        annual_cost_£: cost_£,
+        annual_saving_£: saving_£
+      )
```
iant, line 29: I think this may fail if `agreed_availability_charge` is `nil` as (afaics) `cost_£` & `saving_£` are undefined.

app/services/costs/tariff_information_service.rb
```diff
@@ -0,0 +1,97 @@
...
+      #find missing billing period and extract earliest and latest dates
+      @missing_periods ||= find_billing_periods(false).map do |period_block|
+        [period_block.first[0], period_block.last[0]]
+      end
+    end
+
+    private
+
+    def calculate_percent_real
+      count = billing_periods.values.count { |missing| missing }
```
iant, line 49: is `{ |missing| missing }` needed here?  (`count = billing_periods.values.count` should suffice if think)

iant (approved):
looks good to me @ldodds couple of small suggestions/nitpicks comments

app/services/costs/agreed_supply_capacity_service.rb
ldodds: Have refactored, this one is a bit of a work in progress

app/services/costs/tariff_information_service.rb
ldodds: just count would count all values, this current code is only counting entries which have a value of "true"

PR #467, opened 2023-03-08
Fix accounting cost chart

ldodds:
Fixes an issue that is affecting the accounting cost charts. 

Currently, if a school has solar then the chart framework is selecting the wrong meter to use to build the chart. It's picking the aggregate electricity meter which includes the self-consumption. This usage shouldn't be included in the calculations.

This fix includes several changes:

1. Fix Chart Configuration

* Updated configuration of affected charts to add `meter_definition:  :allelectricity_unmodified`

The following charts were updated:

```
electricity_cost_comparison_last_2_years_accounting
electricity_cost_comparison_last_2_years_accounting_breakdown
electricity_cost_1_year_accounting_breakdown
electricity_1_year_intraday_accounting_breakdown
```

While there is an inheritance hierarchy here, to avoid any ambiguity, the meter definitions have been set on each chart.

The majority of the affected charts were identified by running

```
bundle exec ruby script/tools/chart_inheritance.rb -n electricity_cost_comparison_last_2_years --parents
```

The final one was found by manual review.

2. Fix Series Manager

The `AccountingCost` class in the SeriesManager is responsible for creating the series names, finding the right meter, and adding up the components. In order for the above fix to work, several other changes were required:

* Update `day_breakdown` to ensure that bill component costs are summed correctly
* Update `series_names` to pull list of bill components from aggregate meter, as sub meter tariffs have an empty list of bill components
* Overide `determine_meter` to add custom logic for finding the right meter when the application is requesting a chart for a specific `mpan_mprn`. This overrides the `meter_definition` so has to be handled in the series manager code.

See code comments for further notes

These changes have currently been tested with:

- st-nicolas-primary-school - no solar, electricity cost component chart and table match
- west-witney-primary-school - has solar, electricity cost component chart and table now match as expected
- king-james-1-community-academy - both electricity and gas component chart and table are producing the right values
- mallaig-high-school - no solar, with storage heaters. Electricity cost component chart and table match


PhilipTB:
I wonder whether _"Update series_names to pull list of bill components from aggregate meter, as sub meter tariffs have an empty list of bill components"_ is necessary as it should at this stage have a full list of bill components I would hope either as a results of the aggregation process or an on the fly calculation?

Otherwise the changes look good.

lib/dashboard/charting_and_reports/charts/series_data_manager.rb
```diff
@@ -812,8 +812,41 @@ def half_hour_breakdown(date, hhi)
...
+    #But sometimes we are running charts with for a specific real or synthetic
+    #mpan. This is used for per-meter cost charts and, as it happens, the aggregate
+    #cost chart in the application.
+    #
+    #In this case we need to use alternative approach to finding the right meter
+    def determine_meter
+      #Just return super if its a generic meter specification
+      return super if chart_config[:meter_definition].is_a? Symbol
+      #Find the meter
+      meter = @school.meter?(chart_config[:meter_definition], true)
```
iant, line 836: would we need to handle if `@school.meter?(chart_config[:meter_definition], true)` returns `nil` here?  (would it ever return `nil`?)

iant (approved):
:+1: Looks good to me @ldodds   one comment on handling a possible nil meter

lib/dashboard/charting_and_reports/charts/series_data_manager.rb
ldodds: That shouldn't happen as the meter id is provided by the application.

PR #471, opened 2023-03-09
I18n support for Benchmark Configuration - column headings

iant:


ldodds (changes requested):
I think you still need to handle the `column_heading_refers_to_previous_year` and `column_heading_refers_to_last_year` methods, in the benchmark configuration class as they're still checking content of strings which might be translated:

`column_heading.to_s.downcase.include?(COLUMN_HEADINGS[key].to_s.downcase)`

ldodds (approved):

ldodds (approved):

PR #476, opened 2023-03-10
Rework monthly costs service

ldodds:
This PR reworks the monthly costs service, the key changes are:

- remove old approach for calculating monthly costs which calculated costs for every month of meter data, even though only up to 2 years was needed
- adds method for getting cost for single month, range of months, last 12 months and previous 12 months
- correctly implement Analysable
- revises tests to variants of the new methods
- removes the MonthlyMeterCollectionCostsService as its now unneeded. With the latest cost page revisions, we only display data for a meter at a time, so don't need to calculate costs across every meter for a school.


iant (approved):
looks good to me @ldodds

PR #487, opened 2023-03-20
Update out of hours benchmarks

ldodds:
Updates the gas and electricity out of hours benchmarking thresholds.

* Removes unused constructor variable from the out of hours alerts
* Renames the constants used in the constructor, to clarify which is the exemplar vs benchmark
* Revised the value of the benchmark so that exemplar is up to 50%, while benchmark is 60%
* Revises the new service to include a comparison vs `benchmark_school`


lib/dashboard/alerts/common/benchmark_metrics.rb
```diff
@@ -22,14 +20,14 @@ module BenchmarkMetrics
...
 
-  GOOD_OUT_OF_HOURS_USE_PERCENT_GAS = 0.3
-  BAD_OUT_OF_HOURS_USE_PERCENT_GAS = 0.7
+  EXEMPLAR_OUT_OF_HOURS_USE_PERCENT_GAS = 0.5
+  BENCHMARK_OUT_OF_HOURS_USE_PERCENT_GAS = 0.6
 
-  GOOD_OUT_OF_HOURS_USE_PERCENT_STORAGE_HEATER = 0.2
-  BAD_OUT_OF_HOURS_USE_PERCENT_STORAGE_HEATER = 0.5
+  EXEMPLAR_OUT_OF_HOURS_USE_PERCENT_STORAGE_HEATER = 0.2
+  BENCHMARK_OUT_OF_HOURS_USE_PERCENT_STORAGE_HEATER = 0.5
```
iant, line 30: perhaps we could we have a comment somewhere on how we arrived at these (revised) exemplar and benchmark percent values (link to a doc etc)?

iant (approved):
looks good to me @ldodds

lib/dashboard/alerts/common/benchmark_metrics.rb
ldodds: Have added a trello link as I did with the other change.

app/classes/usage/annual_usage_category_breakdown.rb
```diff
@@ -77,15 +90,27 @@ def percent_improvement_to_exemplar
...
+      when :gas then BenchmarkMetrics::BENCHMARK_OUT_OF_HOURS_USE_PERCENT_GAS
+      when :storage_heater then BenchmarkMetrics::BENCHMARK_OUT_OF_HOURS_USE_PERCENT_STORAGE_HEATER
+      end
+    end
+
     def exemplar_out_of_hours_use_percent
       # Code adapted from:
       # AlertOutOfHoursElectricityUsage#good_out_of_hours_use_percent = 0.35
       # AlertOutOfHoursGasUsage#good_out_of_hours_use_percent = 0.3
       # AlertStorageHeaterOutOfHours#good_out_of_hours_use_percent = 0.2
```
iant, line 109: sorry @ldodds one extra thing before merging that i missed - the values after these comments (0.35 etc) would be out of date now can be removed.

PR #488, opened 2023-03-20
Revise approach for calculating peak kw benchmark

ldodds:
Revises how we benchmark peak kw usage.

Currently we have a fixed value (`ELECTRICITY_PEAK_USAGE_KW_PER_M2=0.01`) which is used to scale peak kw value based on floor area.

After analysing the data we've decided to switch to using an approach that takes into account both the type of school and the number of pupils (rather than floor area).

This PR:

* Adds new methods to BenchmarkMetric to calculate the peak kw value for different school types, using numbers from the analysis
* Revises the peak kw alert to use the new method for calculating exemplar kw value
* Revises the peak usage benchmarking service to use the new method for calculating exemplar kw value
* Updates that service to support comparing against a `benchmark_school` so that we produce a proper school comparison


iant (approved):
:+1: looks good @ldodds

PR #490, opened 2023-03-21
Fix for community use issue

ldodds:
This PR has a proposed fix for an issue calculating community usage for schools.

A school on EnergySparks has community use period in the morning, but for at least one day the calculation that assigns usage and baseload to community use was producing negative figures.

This is because while the community use in kwh may be set to a minimum of zero, the community baseload usage is still calculated as a portion of the actual baseload.

If the community and community baseload are both set to zero when there is minimal community use the calculations work.


PhilipTB:
Change looks ok to me.

PR #491, opened 2023-03-21
Revise rating logic for AlertPeriodComparisonBase

ldodds:
AlertPeriodComparisonBase has a custom `calculate_rating` method which is intended to stop the alerts from displaying if the cost savings is between +/- £10.

But the method was doing the following:

```
return 10.0 if financial_difference_£.between?(-MINIMUM_DIFFERENCE_FOR_NON_10_RATING_£, MINIMUM_DIFFERENCE_FOR_NON_10_RATING_£)
```

This means that the alert was being assigned the maximum rating for minor cost savings. The net result of which was that the alerts are ranked higher than others, which is the opposite of what is intended. So users are seeing alerts for, e.g. a 74p cost saving. 

The PR revises the code to return nil if the cost saving is within +/- £10. If there's no rating then the application will not try and display the results.

This is done here, rather than marking the alert as not relevant, because the calculations need to be carried out before the assessment can be made.


iant (approved):
looks good to me @ldodds

PR #512, opened 2023-04-05
Translate chart timescales

ldodds:
Ensures that all the time periods used for chart timescales are translatable.

There PR also adds some code comments and additional rspecs for the related classes.

### DIFF
--- a/lib/dashboard/charting_and_reports/charts/chart_timescale_descriptions.rb
+++ b/lib/dashboard/charting_and_reports/charts/chart_timescale_descriptions.rb
@@ -9,7 +9,7 @@ class ChartTimeScaleDescriptions
 
   def self.timescale_name(timescale_symbol) # also used by drilldown
     #using default rather than falling back to :none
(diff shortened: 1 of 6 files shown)


lib/dashboard/charting_and_reports/charts/chart_timescale_descriptions.rb
```diff
@@ -79,10 +63,17 @@ def self.impute_description_from_date_range(date_range)
...
       end
     end
   end
-end
\ No newline at end of file
+
+  def self.known_type?(timescale)
+    #test for :none as fallback behaviour for any code relying on
+    #previous behaviour
+    I18n.t(TIMESCALE_TYPES).key?(timescale) || timescale == :none
```
iant, line 77: an alternative to this could be to look for the specific translation, and return nil if not found, but I'm not sure if it's any more readable than what you have here
```
def self.known_type?(timescale)
  I18n.t("#{TIMESCALE_TYPES}.#{timescale}", default: nil).present? || timescale == :none
end
```

iant (approved):
:+1: looks good to me @ldodds

PR #514, opened 2023-04-06
School dashboard charts skip a week

ldodds:
This PR tries to address the following issues:

1. Charts with a timescale of `:year` or `:up_to_a_year` and a `:week` x_axis skip a whole week of data when you move back a year.
2. Finally, charts that are defined as `:up_to_a_year` don't currently allow a user to navigate over the full range of data. E.g. if a school has 18 months of data, you can only view the first year, even though the chart could display the data

There is also an issue that `:year` or `:up_to_a_year` and a `:week` charts also don't show the latest partial week of data, just the last full week. This will be addressed in a future PR. But requires some thought about to how to properly highlight this on the charts.

The cause of the first problem (week skipping) is that the `XBucketWeek` bucketor adjusts the date range for the chart to the first Sunday of the date range its given. Because that is done separately to the code that creates the range of available chart periods (which is done in the `YearPeriod` and `UpToAYearPeriod` classes) we end up with gaps in the overall date range displayed to the user.

The Period classes should define the range for the chart data, the XBucketors should just decide how to group and label that on the axes. So there shouldn't be any manipulation of the dates.

The second problem (limited navigation) is caused by the code in `ChartManagerTimescaleManipulation` just counting number of years for an `up_to_a_year` chart, rather than the number of periods (as returned by `Holidays.period_cadence`).

The issue of not showing partial weeks is because `XBucketWeek` iterates starting from first Sunday, and then discards the final week of data even if we have 5-6 days worth of data.

The initial fixes included here are to:

- change how the `YearPeriod` calculates ranges, by adding the `move_to_saturday_boundary` parameter to the call to `years_to_date`. This results in a full 52 week periods being returned.
- change how the `UpToAYearPeriod` calculates its ranges. This uses a different approach. The PR adds an additional parameter that will allow the period to be optionally aligned with a saturday

The above changes will only apply to charts with a weekly x_axis.

There seems to be little alert code that now reads data from charts, the remaining code is:

- out of hours usage alerts (see also new service). This is unaffected as it has `nodatebuckets` for `x_axis`
- school week comparison. This uses a couple of different charts, but these have different `timescale` and an `x_axis` of `day`

The `AnnualUsageMeterBreakdownService` is impacted by the change. The new service follows the original advice page logic which read data from the chart. Like the chart this will now use a full 52 week period.

The application code may need to be updated to ensure that the date ranges displayed in subtitles are correct.

The PR also adds new rspec tests for `AggregatorSingleSeries` that allows a charts to be generated without needing a full meter collection. This is used to run several charts with different timescales, meter dates, etc. This confirms that the first produces sensible outputs.

Still to do:

* [x] Decide whether to allow `:up_to_a_year` charts to navigate to the full range of data (there is a commented out change in `ChartManagerTimescaleManipulation`)
* [x] Additional testing / review of Alert and service code that reads data from charts: some of these may end up having different date periods


iant (approved):
:+1: Looks good to me @ldodds

PR #522, opened 2023-04-19
Further easter shutdown benchmark changes

ldodds:
* Change introduction text
* Fix issue when school has bad calendar data
* Change gas, electricity and storage heater tables to just show the `difference_kwh`, `difference_£current` and `difference_co2` variables (adding the latter to the list of benchmark variables)
* Remove redundant columns in electricity usage, add co2 % column for gas/storage
* Rework the total energy usage table to total up just the differences, leave % calculation unchange


iant (approved):
:+1: looks good to me @ldodds

PR #530, opened 2023-04-25
Allow date range in recent usage period calculation service

iant:


iant:
closing this as we'll be removing the two recent usage classes in favour of doing things application side

PR #531, opened 2023-05-03
Better reporting of errors calculating economic costs

ldodds:
If we have misconfigured tariffs then we get errors from the tarrif manager, e.g. `undefined method `[]' for nil:NilClass (NoMethodError)`. This isn't very helpful when attempting to debug in the application as its not clear which meter, date or tariff is the issue.

This PR just adds an exception handler and a new custom exception that adds some additional context (the date, the mpan_mprn, and whether we're attempting to calculate a different tariff).


iant (approved):
:+1:

PR #536, opened 2023-05-04
Refactor analytics to allow prices to be provided more dynamically

iant:


ldodds (commented):
I think these changes are OK to merge, however as I've noted in a few places we're updating code that is marked as deprecated, is unused or is commented out.

My feeling is it would be better to clean these up, probably as part of this PR.

If we do that, then this leaves us with the following uses of these variables:

# Solar Price

* used to calculate £ value of exporrted solar, with existing panels for the advice page
* used to calculate economic benefit of installing solar for advice page and alert   

# Oil Price

Used in a calculation that doesn't seem to be used anywhere

# Electricity Price

* used to calculate cost of a kettle, smartphone, tv, console, for equivalences
* used as default electricity tariff when school has gas only, when calculating cost of point of use water heaters
 
# Gas price

* use to calculate cost of a shower for equivalences

ldodds (approved):

PR #555, opened 2023-06-06
Aggregate data service docs

ldodds:
This PR: 

* adds some documentation around the aggregations services
* standardises a few variable names in the those services
* extracts some of the costs and tariff classes into separate files


app/models/tariffs/generic_accounting_tariff.rb
```diff
@@ -0,0 +1,280 @@
...
+      "#{low_threshold.round(0)} to #{high_threshold.round(0)} kwh"
+    end
+  end
+
+  def check_time_ranges_on_30_minute_boundaries(time_ranges)
+    time_of_days = [time_ranges.map(&:first), time_ranges.map(&:last)].flatten
+    if time_of_days.any?{ |tod| !tod.on_30_minute_interval? }
+      raise_and_log_error(TimeRangesNotOn30MinuteBoundary, "Differential tariff time of day rates not on 30 minute interval #{@mpxn}", time_ranges)
+    end
+  end
```
iant, line 272: maybe an early return? 
```
  def check_time_ranges_on_30_minute_boundaries(time_ranges)
    time_of_days = [time_ranges.map(&:first), time_ranges.map(&:last)].flatten
    return unless time_of_days.any?{ |tod| !tod.on_30_minute_interval? }

    raise_and_log_error(TimeRangesNotOn30MinuteBoundary, "Differential tariff time of day rates not on 30 minute interval #{@mpxn}", time_ranges)
  end
```

iant (approved):
:+1: looks good to me @ldodds 
a few suggestions/nitpicks but looks good to me as-is.

ldodds:
Thanks @iant for the moment I'm just extracting some of the files to support later refactoring which we can start next week I think.

PR #556, opened 2023-06-08
Add new solar generation summary benchmark

ldodds:
To support testing of the revised solar export / self consumption applications I've added a new admin-only benchmark that summarises the key figures for the solar panel peformance:

- generation
- self consumption
- export
- main consumption
- mains + self consumption


iant (approved):
:+1:

PR #559, opened 2023-06-14
Add kwh reduction as a variable for priority actions

ldodds:
This PR covers updating the alerts to add a potential kwh saving to the common variables used to describe one year cost and co2 savings and capital costs. These variables are used to display the management priorities and we want to add the kwh reduction to that.

* [x] add spec for the `AlertAnalysisBase.set_savings_capital_costs_payback` method that assigns variables as basis for refactoring
* [x] refactor the above method implementation and tidy up the code
* [x] create a new method  `AlertAnalysisBase.assign_commmon_saving_variables` which uses keywords arguments to make it clearer which value is which, then revise the current method to call that
* [x] update all the alerts to use the new base class method, prior to it being removed, simplifying the passed arguments where possible
* [x] revise `assign_commmon_saving_variables` method to support adding an optional `one_year_saving_kwh` argument and a template variable declaration so the value is available to the application 
* [x] update alerts to add the `one_year_saving_kwh` argument where possible

For the last step above, the kwh reductions have been taken from: 

* existing template variables which already provided a kwh saving. In this case the new `one_year_saving_kwh` variable is just an alias for an existing value, giving it a consistent name for use by the application
* OR calculated from other variables already produced in the alert. In the majority of cases the potential kwh saving is calculated and then the cost and £/£current values derived from it, so have just used those

There are a few alerts where a `one_year_saving_kwh` isn't suitable because they're about cost reduction and not energy saving specifically. E.g. a meter consolidation alert or an alert about the reaching the Agreed Supply Capacity limit.

The above changes have been tested through:

* the rspecs that confirm the `AlertAnalysisBase` class performs as expected
* using the test_alerts.rb script and running alerts for the alerts for a range of 20 schools to catch any runtime errors and confirm that variables are being set correctly
* testing via the application, e.g. does a school successfully regenerate and save the new variable as expected

### DIFF
--- a/lib/dashboard/alerts/common/alert_analysis_base.rb
+++ b/lib/dashboard/alerts/common/alert_analysis_base.rb
@@ -265,11 +265,9 @@ class AlertAnalysisBase < ContentBase
 
   def assign_commmon_saving_variables(one_year_saving_£:, capital_cost:, one_year_saving_co2:)
     @one_year_saving_£ = as_range(one_year_saving_£)
-    #TODO one_year_saving_£ can never be nil
     @ten_year_saving_£ = @one_year_saving_£.nil? ? 0.0 : Range.new(@one_year_saving_£.first * 10.0, @one_year_saving_£.last * 10.0)
 
     #PRIORITY
-    #TODO one_year_saving_£ can never be nil
     @average_one_year_saving_£ = @one_year_saving_£.nil? ? 0.0 : ((@one_year_saving_£.first + @one_year_saving_£.last) / 2.0)
     @average_ten_year_saving_£ = @average_one_year_saving_£ * 10.0
 
@@ -279,7 +277,6 @@ class AlertAnalysisBase < ContentBase
 
     @capital_cost = as_range(capital_cost)
     #PRIORITY - dashboard only
-    #TODO capital_cost can never be nil
     @average_capital_cost = @capital_cost.nil? ? 0.0 : ((@capital_cost.first + @capital_cost.last)/ 2.0)
 
     @average_payback_years = @average_one_year_saving_£ == 0.0 ? 0.0 : @average_capital_cost / @average_one_year_saving_£
--- a/lib/dashboard/alerts/common/alert_analysis_base.rb
+++ b/lib/dashboard/alerts/common/alert_analysis_base.rb
@@ -265,9 +265,11 @@ class AlertAnalysisBase < ContentBase
 
   def assign_commmon_saving_variables(one_year_saving_£:, capital_cost:, one_year_saving_co2:)
     @one_year_saving_£ = as_range(one_year_saving_£)
+    #TODO one_year_saving_£ can never be nil
     @ten_year_saving_£ = @one_year_saving_£.nil? ? 0.0 : Range.new(@one_year_saving_£.first * 10.0, @one_year_saving_£.last * 10.0)
 
     #PRIORITY
+    #TODO one_year_saving_£ can never be nil
     @average_one_year_saving_£ = @one_year_saving_£.nil? ? 0.0 : ((@one_year_saving_£.first + @one_year_saving_£.last) / 2.0)
     @average_ten_year_saving_£ = @average_one_year_saving_£ * 10.0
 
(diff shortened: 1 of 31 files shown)


lib/dashboard/alerts/common/alert_analysis_base.rb
```diff
@@ -263,29 +268,25 @@ def add_book_mark_to_base_url(bookmark)
     @help_url = ALERT_HELP_URL + '#' + bookmark
   end
 
-  def calculate_payback_years_deprecated
-    return (0.0..0.0) if one_year_saving_£.nil? || capital_cost.nil? || capital_cost == (0.0..0.0)
-    min_saving = one_year_saving_£.last.nil? ? 0.0 : capital_cost.first / one_year_saving_£.last
-    max_saving = one_year_saving_£.first.nil? ?  0.0 : capital_cost.last / one_year_saving_£.first
-    Range.new(min_saving, max_saving)
-  end
+  def assign_commmon_saving_variables(one_year_saving_kwh: nil, one_year_saving_£:, capital_cost: nil, one_year_saving_co2:)
```
iant, line 271: just to double check - do we also need `nil` default values for the `one_year_saving_£` & `one_year_saving_co2` keywords here?

lib/dashboard/alerts/common/alert_analysis_base.rb
```diff
@@ -263,29 +268,25 @@ def add_book_mark_to_base_url(bookmark)
...
+    @one_year_saving_£ = as_range(one_year_saving_£)
+    @ten_year_saving_£ = @one_year_saving_£.nil? ? 0.0 : Range.new(@one_year_saving_£.first * 10.0, @one_year_saving_£.last * 10.0)
 
-  def set_savings_capital_costs_payback(one_year_saving_£, capital_cost, one_year_saving_co2)
-    one_year_saving_£ = Range.new(one_year_saving_£, one_year_saving_£) if one_year_saving_£.is_a?(Float)
+    @average_one_year_saving_£ = @one_year_saving_£.nil? ? 0.0 : ((@one_year_saving_£.first + @one_year_saving_£.last) / 2.0)
```
iant, line 275: i wonder if it's worth doing an additional check if `@one_year_saving_£` is both present and a (at least) 2 element array of numbers, here?  Something like this perhaps:
```
@average_one_year_saving_£ = (@one_year_saving_£&.first&.is_a?(Numeric) &&  @one_year_saving_£&.last&.is_a?(Numeric)) ? ((@one_year_saving_£.first + @one_year_saving_£.last) / 2.0) : 0.00
```

lib/dashboard/alerts/common/alert_analysis_base.rb
```diff
@@ -263,29 +268,25 @@ def add_book_mark_to_base_url(bookmark)
...
-    @capital_cost = capital_cost
-    @average_capital_cost = capital_cost.nil? ? 0.0 : ((capital_cost.first + capital_cost.last)/ 2.0)
+    @one_year_saving_kwh = one_year_saving_kwh
 
-    @one_year_saving_£ = one_year_saving_£
-    @ten_year_saving_£ = one_year_saving_£.nil? ? 0.0 : Range.new(one_year_saving_£.first * 10.0, one_year_saving_£.last * 10.0)
-    @average_one_year_saving_£ = one_year_saving_£.nil? ? 0.0 : ((one_year_saving_£.first + one_year_saving_£.last) / 2.0)
-    @average_ten_year_saving_£ = @average_one_year_saving_£ * 10.0
+    @capital_cost = as_range(capital_cost)
+    @average_capital_cost = @capital_cost.nil? ? 0.0 : ((@capital_cost.first + @capital_cost.last)/ 2.0)
```
iant, line 284: similar to the above comment.  is it worth checking here if `@capital_cost` is an array of numbers?

iant (approved):
looks good to me @ldodds couple of comments/suggestions but looks good to merge as is.

lib/dashboard/alerts/common/alert_analysis_base.rb
ldodds: The value can only be nil or a Range at this point.

PR #571, opened 2023-07-13
Avoid infinite loop in rescaling of amr data

ldodds:
A bug in the front end means that it's possible to create "rescale" meter attribute that has no start or end date. In the latter case it causes the validation process to loop forever.

This PR makes the correction logic in the analytics more robust, ensuring that the scaling is only done over the available date range. The front-end bug will also get fixed.

The code for the scaling has been broken out into a simple `Rescaler` service, as it makes it easier to test and reduces amount of code in the validation service. This service is then used by the `ValidateAMRData` class. 

My plan, over time, is to migrate more of that correction and validation code into independent, separately testable services.


urbanwide (approved):
Looks good @ldodds 👍

PR #573, opened 2023-07-19
Update rating calculation in peak kw alert

ldodds:
The recent changes to how we calculate the benchmark and exemplar values for the peak kw alert have broken the ratings calculation. The wrong and inconsistent variables are being used.

This PR fixes the rating calculation, to apply the same logic as used in the baseload and annual consumption alerts: generate a rating based on comparing the school's peak kw usage vs that expected of a benchmark and exemplar school with the same number of pupils.

The rating is offset so that a peak kw usage equal to the benchmark gives a rating of 4.0. An exemplar school will score 10.0. A school needing to take action (> benchmark usage) will get a rating of 0.0.

Ratings have been manually checked for a range of schools.


iant (approved):
looks good to me @ldodds

PR #576, opened 2023-07-25
Remove unused alert type and meter attributes

ldodds:
Removes an alert type that we aren't using and won't use in future, as well as the meter attribute it uses for calculations.


iant (approved):
:+1:

PR #577, opened 2023-07-26
Removed unused tariff code

ldodds:
The `MeterTariffManager` includes support for some functionality which is currently unused, this includes:

* ability to specify an "override" accounting tariff that will be used instead of any other tariffs for the meter, e.g. to force DCC tariffs to be ignored
* ability to specific a "merge" tariff that is intended to allow standing charges to be added to other tariffs. (this code is unfinished and contains errors)
* ability to specify an override that will force use of a differential tariff when calculating economic costs for a specific date range

We don't use this currently in the application: there are no attributes configured for any meters in the system.

With the planned rework of the tariff code these will either be unnecessary or will be supported in other ways:

* users can specific tariffs for schools and meters to override those set of a group or system
* tariffs will have a status flag so incorrect / invalid tariffs can be disabled
* DCC tariffs will be editable so that standing charges can be added in
* there will no longer be a distinction between economic and accounting tariffs, there will just be accounting tariffs. So no need to specify an override

This PR:

* Adds rspec tests around all of the core `MeterTariffManager` functionality to support this refactoring and as well as future changes
* Removes all of the code that processes the above meter attributes, along with their definitions
* Tidies up the class to remove some unused methods and to mark a couple of methods that are only used internally as `private`

### DIFF
--- a/app/models/meter_attributes.rb
+++ b/app/models/meter_attributes.rb
@@ -823,22 +823,6 @@ class MeterAttributes
     structure MeterAttributes.generic_accounting_tariff
   end
 
-  class AccountingGenericTariffOverride < MeterAttributeTypes::AttributeBase
-    id :accounting_tariff_generic_override
-    aggregate_over :accounting_tariff_generic_override
-    name 'Accounting tariff override (generic + DCC)'
-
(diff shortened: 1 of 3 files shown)


iant (approved):
:+1:

PR #578, opened 2023-07-28
Add some specs to datetime_helper

ldodds:
This PR just adds some specs and comments to the DateTimeHelper class.

Goal is mostly to highlight the subtle differences in behaviour in two methods used in the tariff code:

* one defines exclusive ranges, that cannot span across midnight and which must use 24:00 as an end range
* one defines inclusive ranges, that span midnight and which does not support 24:00 or 00:00 as an end range. (The latter is the 1st half-hour the day, and the former produces invalid results)


iant (approved):
:+1:

PR #579, opened 2023-07-28
Tidy up solar_pv_profit_loss

ldodds:
Reviewing how we use (or not) solar pv and solar export pv tariffs I noticed that there was some deprecated code in the `SolarPVProfitLoss` class, so this PR just removes that:

- removes constructor arguments not used in class
- removes constants not used anywhere
- remove private methods with a `_deprecated` suffix to their name, plus any methods that only they call
- remove duplicate method declaration, leaving private method


iant (approved):
:+1:

PR #580, opened 2023-08-02
Add tariff holder and created_at to generic tariff attribute

ldodds:
To better distinguish between different tariff holders, I've updated the meter attribute structure to indicate:

- the tariff holder for a specific tariff (meter, if its tied to a specific meter, or school, school group, site settings to match hierarchy)
- a created_at attribute to support disambiguation between potentially overlapping tariffs


iant (approved):
:+1:

PR #581, opened 2023-08-02
Create new Generic Tariff Manager

ldodds:
This PR creates a new tariff manager implementation (`GenericTariffManager`) which is intended to be a drop-in replacement for the current implementation (`MeterTariffManager`).

The new version works solely from `accounting_tariff_generic` meter attributes, as supplied by the new `EnergyTariff` model in the application. This type of meter attribute structure will cover all types of tariff from now on.

The new implementation does not distinguish between economic and accounting tariffs. Instead it just calculates economic costs from the full accounting tariffs, by ignoring the standard charges and other costs.

It also provides more flexibility in creating and managing tariffs by supporting a full hierarchy of tariffs from system, school group, school and meter. These tariffs are sorted into different buckets (via the `tariff_holder` attribute) and are then searched in order to find the applicable tariff for a given day. This ensures that overrides at each level are processed correctly.

It relaxes requirements around overlapping/contiguous tariffs to give flexibility when editing and be more tolerant of errors.

This version does not support weekend/weekday tariffs at present. We don't have any of those currently, so handling of those needs some rework across the system.

Switching to use the new tariff manage is done via the `FEATURE_FLAG_USE_NEW_ENERGY_TARIFF` environment variable.

Due to the removal of economic tariffs, there isn't a direct equivalent of some code, e.g. checking whether tariffs change over time. Sensible defaults have been implemented pending further refactoring of the tariff code.

The PR adds some additional tests around some classes and also removes some redundant / unused code.

In future PRs we can simplify things further:

* simplify the MeterTariff class hierarchy, to just a single AccountingTariff class with subclasses for more esoteric tariffs (e.g. weekend/weekday)
* remove the older MeterTariffManager implementation
* reorganise the classes into more consistent modules

In addition to the provided specs, the changes have been manually tested against https://github.com/Energy-Sparks/energy-sparks/pull/2948 which ensures that the EnergyTariffs are provided to the application in the right structure and all of the older accounting and economic tariffs are ignored when aggregating a school. All schools have been aggregated without errors (except for existing errors in the manually setup meter attributes). Tariff calculations for charts and advice pages also work correctly.

### DIFF
--- a/app/services/tariffs/generic_tariff_manager.rb
+++ b/app/services/tariffs/generic_tariff_manager.rb
@@ -203,48 +203,47 @@ class GenericTariffManager
         raise "Unknown tariff holder type"
       end
     end
+  end
 
-    # tariffs for new SMETS2 meters are often setup several days after
-    # kWh data has started recording, the earlier kWh readings therefore
-    # have no DCC tariff and default to default accounting tariffs
-    # in this circumstance, unless overridden backdate the existing DCC tariff
-    # to the start of the meter readings, so the default is no longer used
-    #
-    # NOTE: as n3rgy no longer hold archived tariffs, then we'll only ever have
-    # the tariffs from the point that we begin loading data. So this may be more
-    # common than it was before
-    #
-    # TODO: this could be done in the application. When the DCC tariffs or readings are loaded for a
-    # meter, the start date of the tariff could be adjusted once. Or the adjustment could
-    # happen when the data is loaded and past to the analytics.
-    def backdate_dcc_tariffs(meter)
-      return if @meter_tariffs.empty? || dcc_tariffs.empty?
+  # tariffs for new SMETS2 meters are often setup several days after
+  # kWh data has started recording, the earlier kWh readings therefore
+  # have no DCC tariff and default to default accounting tariffs
+  # in this circumstance, unless overridden backdate the existing DCC tariff
+  # to the start of the meter readings, so the default is no longer used
+  #
+  # NOTE: as n3rgy no longer hold archived tariffs, then we'll only ever have
+  # the tariffs from the point that we begin loading data. So this may be more
+  # common than it was before
+  #
+  # TODO: this could be done in the application. When the DCC tariffs or readings are loaded for a
+  # meter, the start date of the tariff could be adjusted once. Or the adjustment could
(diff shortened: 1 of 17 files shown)


app/services/tariffs/generic_tariff_manager.rb
```diff
@@ -0,0 +1,255 @@
...
+  end
+
+  #find most recently created tariff, force nil timestamps to sort last
+  def most_recent(found_tariffs)
+    found_tariffs.max_by do |t|
+      t.tariff[:created_at].nil? ? MeterTariff::MIN_DEFAULT_START_DATE : t.tariff[:created_at]
+    end
+  end
+
+
```
iant, line 149: remove empty line here (l149) and below (164)?

lib/dashboard/aggregation/amr_data.rb
```diff
@@ -35,14 +35,14 @@ def set_tariffs(meter)
 
   def set_economic_tariff(meter)
     logger.info "Creating an economic costs in amr_meter #{meter.mpan_mprn} #{meter.fuel_type}"
-    @economic_tariff = EconomicCostsParameterised.create_costs(meter)
+    @economic_tariff = EconomicCostsParameterised.new(meter)
     # @current_economic_tariff = @economic_tariff
```
iant, line 39: can this commented line be removed now?

app/services/tariffs/generic_tariff_manager.rb
```diff
@@ -0,0 +1,255 @@
...
+    prev_tariff.start_date == MeterTariff::MIN_DEFAULT_START_DATE ? nil : prev_tariff.start_date
+  end
+
+  # Find all the dates when the tariff changes within the period
+  #
+  # AlertEconomicTariffCalculations
+  # Costs::EconomicTariffsChangeCaveatsService
+  def tariff_change_dates_in_period(start_date  = @meter.amr_data.start_date, end_date = @meter.amr_data.end_date)
+    list_of_tariffs = find_tariffs_between_dates(start_date, end_date)
+    return list_of_tariffs.map{ |t| t.end_date + 1 }
```
iant, line 93: return not needed here

iant (approved):
:+1: looks good to me @ldodds couple of minor clean up bits (empty line and commented text)

PR #583, opened 2023-08-08
Fix bug with periods not on half-hourly boundaries

ldodds:
Fixes an issue with periods that aren't on exact half-hourly boundaries.

For the start of the half-hourly period, the code was returning the time within the period, rather than the time remaining.

This potentially impacts all calculations that involve school opening hours: our system defaults are 8.50am to 15:20. The bug would have caused the 8.50am range to have been treated as 20 rather than 10 minutes within the 8.30-9.00am range. See updates to usage breakdown service for indication of impacts.

Potentially also impacts community use times, depending on data provided by schools.


iant (approved):
:+1:

PR #584, opened 2023-08-08
Additional tariff code tests and refactoring

ldodds:
This is a further refactoring of the tariff code to simplify some of the code and add more rspec coverage:

* Adds keyword arguments to OneDaysCostData
* Revises the tariff manager and tariff interfaces to directly create instances of OneDaysCostData, rather than first creating a Hash then copying it into the object
* Use the above to extract out some common code within the GenericAccountingTariff `.costs` and `.economic_costs` methods


lib/dashboard/aggregation/costs.rb
```diff
@@ -40,6 +41,36 @@ class CostsBase < HalfHourlyData
...
+    (combined_start_date..combined_end_date).each do |date|
+      list_of_meters_on_date = list_of_meters.select { |m| date >= m.amr_data.start_date && date <= m.amr_data.end_date }
+
+      #If provide block returns true then we'll skip the cost calculation
+      #for this date
+      skip = false
+      if block_given?
+        skip = yield(date, list_of_meters_on_date)
+      end
+      next if skip
```
iant, line 65: would it work to move lines 61 to 65 into 1 line?
```
next if block_given? && yield(date, list_of_meters_on_date)
```

iant (approved):
This looks good to me @ldodds :+1:

PR #588, opened 2023-08-10
Remove unneeded scan of AMR data

ldodds:
When we validate data there are some status markers added to the validated readings:

- LGAP which means the first reading after a long gap of missing readings. 
- FIXS which indicates we've forced the system to only look at a fixed start date
- FIXE the above, but for end date

Whenever those are set the validation code also calls `set_start_date` or `set_end_date` on the `AmrData` class. This ends up removing all of the readings before or after that. When this is passed back to the application for storage we just end up storing that range of data.

There has been a check in the aggregation service for sometime that looks for the LGAP, FIXS and FIXE markers and then calls `set_start_date` and `set_end_date` again. This is based on the assumption that the application is storing data before and after those dates and so the dates have to be re-applied based on the markers.

That's not actually the case though. The application only stores the validated data returned by the analytics. Anything outside of the range that might have existed is removed.

So the `set_long_gap_boundary_on_all_meters` check is doing an unnecessary scan of all AMR data across all the meters for a school. It does find the LGAP/FIXS/FIXE markers but just ends up setting the same data. 

This PR removes that code as its redundant. I've added some comments about the application behaviour to the validation service where it sets the markers.

### DIFF
--- a/app/services/aggregation_service.rb
+++ b/app/services/aggregation_service.rb
@@ -70,9 +70,6 @@ class AggregateDataService
   def aggregate_heat_and_electricity_meters
     log 'Aggregate Meters'
     bm = Benchmark.realtime do
-      #adjust meter start/end dates if needed
-      set_long_gap_boundary_on_all_meters
-
       #aggregate heat meters
       aggregate_heat_meters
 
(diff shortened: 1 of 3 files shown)


iant (approved):
:+1:

PR #598, opened 2023-09-06
Fix divide by zero error generating electricity progress

iant:


lib/dashboard/modelling/targeting and tracking/target_dates.rb
```diff
@@ -175,6 +175,9 @@ def annual_kwh_estimate_helpful?
   def percentage_synthetic_data_in_date_range(start_date, end_date)
     total_days = end_date - start_date + 1
     synthetic_days = (start_date..end_date).count{ |date| date < benchmark_start_date && date >= synthetic_benchmark_start_date }
+    
+    return 0.0 if total_days.nil? || total_days == 0.0
```
urbanwide, line 179: Just pondering why you used total_days == 0.0 rather than just 0 here? Also thinking you could add the checks you've put on the Trello card to a test if it's not too much hassle anyway...

urbanwide (approved):
Looks good @iant 👍

PR #605, opened 2023-10-03
Only calculate optimum not full range of scenarios

ldodds:
Some small optimisations to the solar benefits estimator alert.

The alert calculates the optimum kwp for installing solar. It also used to generate an additional range of scenarios that was used to populate the old analysis pages. 

This functionality has been moved to the `PotentialBenefitsEstimatorService` class which now produces the table for the advice page. So this PR streamlines the alert to remove the calculations no longer required here. 

Due to the use of the `Minimiser` it may still end up doing up to 10 simulations to calculate the optimum installation size, so still quite costly. But these changes will reduce the overall number of calculations. The larger the floor area in the school, the more the benefit as the `max_possible_kwp` method was used to decide how many extra scenarios to calculate.


iant (approved):
:+1:

PR #606, opened 2023-10-03
Add AnnualUsageBreakdownService.annual_out_of_hours_kwh

ldodds:
The `AnnualUsageBreakdownService` is used to calculate the kwh, co2 and costs for different categories of the school week: out of hours, open, closed, community usage.

The service was built for the new advice pages and by default returns the calculations for all three variables across all time periods.

However we are currently using the service in the `OutOfHoursUsageBenchmarkGenerator` in the application to just extract the out of hours kwh and total kwh for the year. This means that we incur costs of calculating additional metrics that we don't need.

This PR adds a new method for just returning those kwh values. This avoids having to calculate the co2 and cost values, which involve doing a lot of unnecessary aggregation of the data (via the charting framework). 

This adds up to avoiding running 3 extra charts/calculations per fuel type for each school.

### DIFF
--- a/app/services/usage/annual_usage_breakdown_service.rb
+++ b/app/services/usage/annual_usage_breakdown_service.rb
@@ -11,6 +11,12 @@ module Usage
       @fuel_type = fuel_type
     end
 
+    def annual_out_of_hours_kwh
+      build_usage_category_usage_metrics!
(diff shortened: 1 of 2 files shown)


iant (approved):
:+1:

PR #607, opened 2023-10-09
Cache kwh date range

ldodds:
During running of alerts and other calculations we make repeated calls to `kwh_date_range` often with the same dates and types.

This PR adds a cache of precalculated values to the AMRData class to remove the need for repeated calculations. This is primarily intended to reduce CPU overheads of recalculating values rather than increase speed.


ldodds:
May re-open this later.

PR #609, opened 2023-10-11
Add notify_aggregation_complete!

ldodds:
This PR implements an alternative approach to the "date bound schedule data" feature which was implemented in the application but never fully released.

We have increasingly long archives of temperature and solar data for some locations, and grid carbon intensity for the whole UK. This schedule data is passed to every meter collection along with the meter data.

However, in general, we don't need all this data. For some target meter and estimated annual usage calculations we need just over a year of temperature data before the earliest meter date. Otherwise we need nothing earlier than that date. For most schools this will only be a couple of years worth of data.

One approach to limiting the amount of data used is to only build the meter collection with a limited dataset. We've tried this on the application and confirmed that the meter collection sizes in the cache are much smaller, so this is worth doing. However with recent changes to the regeneration process we need to revisit and change that code

Another approach is to delegate to the meter collection to prune back any data it doesn't need, following completion of the aggregation process. This works regardless of whether we're starting from validated or unvalidated meter data.

This PR implements this second approach. It is slightly simpler in overall implementation and already builds on a post-aggregation process of settings flags across the meter data to enable caching, etc. 

It also gives us a foundation for potentially doing other clean-up or optimisations if required.

### DIFF
--- a/app/models/meter_collection.rb
+++ b/app/models/meter_collection.rb
@@ -482,4 +482,35 @@ class MeterCollection
   def is_school_usually_open?(_date, time_of_day)
     time_of_day >= @cached_open_time && time_of_day < @cached_close_time
   end
+
+  # Notify meter collection that aggregation process is over.
+  # Allows for any post aggregation clean-up to be carried out.
+  def notify_aggregation_complete!
+    clean_up_schedule_data!
+
+    # Set flags on each meter to indicate aggregation process has
+    # been completed.
+    #
+    # allows parameterised carbon/cost objects to cache data post
+    # aggregation, reducing memory footprint in front end cache prior to this
+    # while maintaining charting performance once out of cache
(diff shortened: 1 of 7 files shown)


app/models/meter_collection.rb
```diff
@@ -482,4 +482,35 @@ def meter_attribute_types
...
+    all_meters.each do |meter|
+      meter.amr_data.set_post_aggregation_state
+    end
+  end
+
+  #Clip the schedule data to the earliest date that we need for charting or
+  #subsequent analysis.
+  private def clean_up_schedule_data!
+    earliest_date = first_combined_meter_date
+    unless earliest_date.nil?
```
iant, line 506: could do an early return if here 
`return if earliest_date.nil?`

iant (approved):
👍 : looks good to me @ldodds

PR #611, opened 2023-10-13
Remove use of old tariff code

ldodds:
The main purpose of this PR is to remove the use of the feature flag controlling whether to use the old or new tariff manager code. This will allow us to remove some additional unused code and support for development.

However, because our specs are based on actual school data, which is present in the project as meter collections, after removing this flag I've also had to update the meter collections for those schools. 

This means that every test that relies on those meter collections also needed to be updated. Since we added the meter collections, the schools have had new or updated data loaded, as well as configuration and tariff updates. So a lot of tests have had to be changed.

The decision to use real-world data was due to there being little/no automated tests in the project. So adding specs using real data, then refactoring around those, was a sensible decision for building the new interfaces. But longer term it has added overheads like this whenever we need to update the data.

The intention is to steadily replace these specs with:

- low level specs that use data from FactoryBot or other synthetic data to confirm that calculations work as expected
- change the service specs to use check that calculations are called and data is returned in the correct structures

This will take some time, so for now I've just taken the hit.

Beyond removing use of the old `MeterTariff` manager there are no functional changes in this spec the other changes are:

- adding the updated meter collections
- revising the specs to confirm the new values that are being returned due to new/changed data and tariffs
- start removing the `round_to_two_digits` custom matcher in favour of the rspec built in.

### DIFF
--- a/spec/app/services/heating/heating_start_time_service_spec.rb
+++ b/spec/app/services/heating/heating_start_time_service_spec.rb
@@ -2,7 +2,7 @@ require 'spec_helper'
 
 describe Heating::HeatingStartTimeService, type: :service do
 
-  let(:asof_date)      { Date.new(2022, 2, 1) }
+  let(:asof_date)      { Date.new(2023, 10, 11) }
   let(:service)        { Heating::HeatingStartTimeService.new(@acme_academy, asof_date)}
 
   #using before(:all) here to avoid slow loading of YAML and then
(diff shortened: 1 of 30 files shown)


spec/app/services/heating/heating_start_time_service_spec.rb
```diff
@@ -31,32 +31,13 @@
...
       first_day = start_times.days[0]
-      expect(first_day.date).to eq Date.new(2022,1,25)
-      expect(first_day.heating_start_time).to eq TimeOfDay.new(5,0)
-      expect(first_day.recommended_time).to eq TimeOfDay.new(0,0)
-      expect(first_day.temperature).to eq -0.3
-      expect(first_day.saving.kwh).to eq 0
-      expect(first_day.saving.£).to eq 0
-      expect(first_day.saving.co2).to eq 0
-
-      last_day = start_times.days.last
```
urbanwide, line 42: Just checking if you meant to remove the spec for last_day here?

urbanwide:
Looks good @ldodds - just one question - see inline 👍

urbanwide (approved):

spec/app/services/heating/heating_start_time_service_spec.rb
ldodds: I did, because I wasn't sure it was adding much. We're already checking there's the right number of days and that there are values for the first day.

urbanwide: Ok great, that's all good then 👍

PR #612, opened 2023-10-13
Simplify tariff code

ldodds:
Branched from #611 which will need to be reviewed and merged first.

This PR simplifies the tariff code. Previously we had several different tariff classes but since the change to the new energy tariff system we have only been using the `GenericAccountingTariff` and `GenericTariffManager` classes.

So the changes here are to:

* remove the old tariff manager implementation and some related classes that were no longer in use
* remove the `AccountingTariff`, `EconomicTariff`, `EconomicTariffChangeOverTime` classes
* collapse the hierarchy of `MeterTariff` < `AccountingTariff` < `GenericAccountingTariff` so that we just have `GenericAccountingTariff`. The class is larger, but is now self-contained
* tidy up the public interface of `GenericAccountingTariff`
* Remove the `AlertDifferentialTariffOpportunity` class which is not enabled on live and won't be used. This was never updated to the newer tariff structure so was already broken.

Further work is required to remove the older meter attributes that describe the older tariff structures, but that will be done in a separate PR.


app/models/tariffs/generic_accounting_tariff.rb
```diff
@@ -289,7 +391,7 @@ def tiered_rate_sub_type?(type)
   end
 
   def tier_description(tier_name, low_threshold, high_threshold, rate_config)
-    time_range = MeterTariff.format_time_range(rate_config)
+    time_range = format_time_range(rate_config)
```
urbanwide, line 394: Just wondered if calling format_time_range without either the classname or self.class.format_time_range will work here. Noticed on line 324 it's called with the classname that's all.

urbanwide (approved):
This looks like a really good refactor @ldodds. See one comment about how format_time_range is called. One for the future: I wonder if GenericAccountingTariff could even be renamed as just Tariff, if that's the only one left. I don't fully understand the implications of that though, so just an idea!

app/models/tariffs/generic_accounting_tariff.rb
ldodds: Yep, I missed that. Thanks

PR #613, opened 2023-10-19
Fix earliest date calculation

ldodds:
Testing the recent changes to limit the schedule data I realised there was a bug in the date calculation.

I was using the `first_combined_meter_date` which doesn't actually calculate the earliest meter date range for all meters. It calculates the most recent start date for aggregated meters only.

I've added an `earliest_start_date` method which will return the earliest start date for any meter in the collection. This will ensure that the schedule data covers all of the amr data time series we have, plus at least a year or so extra for temperature data. We only need this when doing meter specific calculations and probably only then for gas meters, so opportunity to optimise further.

I've also tweaked the `set_start_date` and `set_end_date` methods in the HalfHourly data class. It will now ignore the set if the date is before the earliest date, or after the latest. This avoids scenario where we might set that date and end up with "gaps" in the data where the object believes its start date is before what is actually held in the internal hash.

Whilst this is a basic change in behaviour, I've reviewed where this method is called and don't think it will impact anything. We're only using it during validation (where we're always truncating data) and during limiting the schedule data.

### DIFF
--- a/app/models/meter_collection.rb
+++ b/app/models/meter_collection.rb
@@ -163,6 +163,10 @@ class MeterCollection
     @calculated_floor_area_pupil_numbers ||= FloorAreaPupilNumbers.new(@floor_area, @number_of_pupils, pseudo_meter_attributes(:school_level_data))
   end
 
+  def earliest_meter_date
+    all_meters.map{|meter| meter.amr_data.start_date }.min
+  end
+
   def first_combined_meter_date
     all_aggregate_meters.map{ |meter| meter.amr_data.start_date }.max
   end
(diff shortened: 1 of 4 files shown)


app/models/meter_collection.rb
```diff
@@ -163,6 +163,10 @@ def calculate_floor_area_number_of_pupils
     @calculated_floor_area_pupil_numbers ||= FloorAreaPupilNumbers.new(@floor_area, @number_of_pupils, pseudo_meter_attributes(:school_level_data))
   end
 
+  def earliest_meter_date
+    all_meters.map{|meter| meter.amr_data.start_date }.min
+  end
+
   def first_combined_meter_date
```
iant, line 170: can this method be removed now?

iant (approved):
:+1: looks good to me @ldodds 
on comment/question on the removal of `first_combined_meter_date`

app/models/meter_collection.rb
ldodds: It can probably be removed, but I think I'll do a separate PR to clean up the class

PR #614, opened 2023-10-20
Allow an extra year of solar data

ldodds:
When removing unneeded schedule data we need to also leave a year of solar data, not just temperature data. This is used to calculate annual estimates for schools that need it.


iant (approved):
:+1: this looks good to me @ldodds

PR #615, opened 2023-10-20
Tidy and reorganise models

ldodds:
This PR does a general tidy-up of some of the code in the `app/models` folder:

- moves around some files into the right directories
- creates separate files for classes that are currently bundled into the same file
- moves methods from `MeterCollection` to `Dashboard::School` where they are more appropriate
- refactors MeterCollection to avoid copying around values from the school, instead just calling methods on the school
- removing old, unused methods
- remove some old temporary code used to calculate open/close times and which is no longer being used
- removed checks for nil schools, as we always construct a MeterCollection with a `Dashboard::School`
- moving some other code used only in test classes into those files

For the moment I've not refactored any other methods and am relying on the existing test suite to ensure that nothing is broken.


app/models/dashboard/school.rb
```diff
@@ -52,6 +52,34 @@ def initialize(
...
+      return nil if @location.nil?
+      @location[1].to_f
+    end
+
+    def activation_date
+      return nil if @activation_date.nil?
+      # the time is passed in as an active_support Time and not a ruby Time
+      # from the front end, so can't be used directly, the utc field needs to be accessed
+      # instead
+      t = @activation_date.respond_to?(:utc) ? @activation_date.utc : @activation_date
```
iant, line 70: prob not something for this pr but maybe this could be extracted to a (private?) method?  something like:
```
def usable_date_from(date)
  # the time is passed in as an active_support Time and not a ruby Time
  # from the front end, so can't be used directly, the utc field needs to be accessed
  # instead
  t = date.respond_to?(:utc) ? date.utc : date
  Date.new(t.year, t.month, t.day)
end
```
with activation and creation date methods then reading
```
def activation_date
  return nil if @activation_date.nil?
  
  usable_date_from(@activation_date)
end
```
```
def creation_date
  return nil if @created_at.nil?

  usable_date_from(@created_at)
end
```

iant (approved):
:+1: this looks good to me @ldodds

app/models/dashboard/school.rb
ldodds: I'm going to do some future PRs to tidy further including running Rubocop over methods, so will do this then.

PR #617, opened 2023-10-23
Change the baseload calculation for schools using sheffield solar data

ldodds:
We use an alternate method of calculating baseload when we are using Sheffield solar PV data to model a school's solar panels. This is to avoid problems with the early morning baseload being underestimated due to the Sheffield data not always matching the school's actual generation (and hence consumption) early in the morning. 

So rather than using a "statistical" method of the lowest 8 half-hourly periods, we use an alternate approach of looking at the usage between 8.30pm and midnight.

However for schools that are open in the evenings, we are seeing signs that the alternate calculation isn't performing well either as some schools are open at the start of that period.

So the primary goal of this PR is to add a better alternative which looks at the periods before and after midnight (10pm to midnight, midnight to 2am).

To get there, I have:

- moved the `ElectricityBaseloadAnalysis` class to be in the `Baseload` module and updated all references, which required adding some requires to some of the alerts
- renamed that class to `BaseloadAnalysis`
- refactored some of its methods to make them private and tidy them up using some recommendations from rubocop (run manually)
- extracted the code that implements the current baseload calculations from the `AMRData` class into new subclasses of `BaseloadCalculator`. These calculators are responsible for calculating the baseload for a specific day. The base class provides a factory method for creating the right calculator based on whether we are using Sheffield solar data
- the `AMRData` class now delegates to instances of `BaseloadCalculator`, which simplifies the code somewhat
- I've then added a third type of calculator `AroundMidnightBaseloadCalculator` which can be substituted for the existing `OvernightBaseloadCalculator` via a feature flag.
- I've added tests for the majority of these changes. The existing service specs help check that nothing more fundamental is broken

The rest of the analytics can later be refactored to create a `BaseloadCalculator` directly. This will help to reduce the public interface of the `AMRData` class which is too large and covers a range of different calculations.


urbanwide (approved):
Looks good @ldodds 👍

PR #619, opened 2023-10-27
Clean up meter attribute configuration and filter list provided to the application editor

ldodds:
Some general tidying of the meter attribute configuration code.

* Remove some unused attributes, configuration marked as deprecated and commented out code
* Remove the declarations of the old style tariffs as we are no longer using those
* Tidy up the names of the attributes, to give them consistent prefixes. The names are used to group the attributes in the admin editor in the application
* Add support to `AttributeBase` to allow an attribute configuration to be marked as "internal only".
* Filter the list of attribute configurations provided to the application to exclude those that are internal only
* Marked up a number of attribute configurations as "internal only" so they don't show up for admins.

The "internal only" attributes are used as data structures to pass data between the front and back end, but shouldn't be created by admins as there are dedicated editors in the application. So this helps reduce clutter in the interface.

### DIFF
--- a/app/models/meter_attribute_types.rb
+++ b/app/models/meter_attribute_types.rb
@@ -20,7 +20,7 @@ module MeterAttributeTypes
   ].freeze
 
   class AttributeBase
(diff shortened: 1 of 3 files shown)


iant (approved):
:+1: looks good to me @ldodds

PR #624, opened 2023-11-02
Fix gaps in community use

ldodds:
When a school has "community use" periods configured we allocate a portion of their usage to this usage, separate to the usual school day open/closed or holiday periods. 

This is done by assigning all, or a percentage of the usage in a half-hourly period to either a school day open, closed, community baseload or community use category.

There is an issue with how the allocation works when the usage during the half-hourly period is lower than the baseload for the day. Currently this results in the usage being allocated to normal school usage, and not the baseload. The bug looks to have been a side effect of fixing a previous issue.

As there were no existing tests around the feature this PR adds a comprehensive set of tests and refactors the relevant code, in addition to just fixing the bug.

A brief summary of the changes:

* refactors and tidies the code associated with community use, removing some older unused code and test scripts
* run rubocop over the relevant files to standardise formatting and fix issues
* change visibility of methods to simplify refactoring
* extracts the open/close time classes into separate files
* refactors the `OpenCloseTimes` class so that it can be directly created with opening and closing time period specifications, and not just via the hashes provided by the application. This makes it much easier to test the class in isolation
* refactors the `AMRDataCommunityOpenCloseBreakdown` class so it doesnt directly use the meter collection associated with the meter its analysing. The existing attributes were sufficient, so this reduces a direct dependency and again makes it easier to test the class in isolation
* renamed `AMRDataCommunityOpenCloseBreakdown` to `CommunityUseBreakdown`
* adds a comprehensive set of specs for `CommunityUseBreakdown` so its core behaviour is fully tested

...and then finally adds a spec to isolate the issue with allocating community usage during periods of usage below baseload, along with a fix. The fix also slightly optimises the calculation so that periods which are 100% community usage use a simpler set of calculations.


urbanwide (approved):
Looks good @ldodds 👍

PR #625, opened 2023-11-13
Ensure usage is properly allocated across all half-hourly splits

ldodds:
My [recent fix to the community use allocation](https://github.com/Energy-Sparks/energy-sparks_analytics/pull/624) seems to have introduced or -- judging from the comments and exception in the code -- reintroduced a bug.

The code was not properly dealing with a couple of scenarios:

- when a half-hourly period was split between community use and school day open, with no time when school was closed, e.g. school closes at 16:15 then community use period starts at 16:15
- when the half-hourly period was split across 3 different usage types, e.g. school is open until 16:10, is closed for 5 minutes and the community use time starts at 16:20

...but only when the usage in those half-hourly periods was below the baseload.

This was only happening for 4 schools during testing as a specific combination of opening hours and usage was required to expose the bug.

This further rework and the additional tests should now ensure that we've got coverage across all possible combinations of half-hourly splits, e.g for a half-hourly period the logic is now checking:

- is completely closed
- if school is open
- if period is entirely community use
- if period is split between open and community use
- if period is split between all three

...and then the tests check for periods where usage is above, at and below the baseload

### DIFF
--- a/spec/lib/dashboard/aggregation/community_use_breakdown_spec.rb
+++ b/spec/lib/dashboard/aggregation/community_use_breakdown_spec.rb
@@ -35,7 +35,7 @@ describe CommunityUseBreakdown do
           expect(days_kwh_x48.values.all?{ |v| v.size == 48 })
         end
 
-        it 'returns same total as the amr_data class' do
+        it 'should return the same total as the amr_data class' do
           expect(days_kwh_x48.values.flatten.sum).to be_within(0.0001).of( meter.amr_data.days_kwh_x48(day).sum )
         end
       end
@@ -52,7 +52,7 @@ describe CommunityUseBreakdown do
           expect(days_kwh_x48.keys).to match_array([:school_day_closed, :school_day_open, :community, :community_baseload])
         end
 
-        it 'returns same total as the amr_data class' do
+        it 'should return the same total as the amr_data class' do
           expect(days_kwh_x48.values.flatten.sum).to be_within(0.0001).of( meter.amr_data.days_kwh_x48(day).sum )
         end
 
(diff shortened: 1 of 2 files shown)


spec/lib/dashboard/aggregation/community_use_breakdown_spec.rb
```diff
@@ -83,7 +83,93 @@
...
+
+        let(:school_times) do
+          [{day: :monday, usage_type: :school_day, opening_time: TimeOfDay.new(7,30), closing_time: TimeOfDay.new(16,10), calendar_period: :term_times}]
+        end
+
+        let(:community_use_times)          do
+          [{day: :monday, usage_type: :community_use, opening_time: TimeOfDay.new(16,15), closing_time: TimeOfDay.new(21,00), calendar_period: :term_times}]
+        end
+
+        it 'should return the expected breakdown across the day' do
```
urbanwide, line 134: Very small style thing - and I wonder if Rubocop should be picking these up: https://rspec.rubystyle.guide/#should-in-example-docstrings

`it 'returns the expected breakdown across the day'`
rather than:
`it 'should return the expected breakdown across the day'`

There a few of these in this file that would need updating.

urbanwide (approved):
Looks good @ldodds - just a small rspec style comment 👍

spec/lib/dashboard/aggregation/community_use_breakdown_spec.rb
ldodds: I think these will get caught by rubcop, I need to merge the other PR that ensures its running

PR #631, opened 2023-11-17
Update average school data

ldodds:
I've run the script to regenerate the `average_school_data.rb` file which contains some hard-coded values that are used to generate some charts. The values are produced using some analysis in the `generate_average_school.rb` script.

I've also removed an unused file produced by a prototype script. I've updated the prototype with some notes for the moment. All of these needs to be re-engineered so we can more easily update the data.


urbanwide (approved):
👍

PR #632, opened 2023-11-17
Update annual usage benchmark values

ldodds:
This PR updates the per-pupil values we use to estimate usage for "well managed" and "exemplar" schools. These are based on updated review of actual school data.

The change also includes a new special case for special schools. These school typically have higher energy consumption due to extra equipment, etc. We now have enough of these schools on Energy Sparks to make a more accurate and fairer assessment of their usage.

The changes to the benchmarks impact the specs for the `Usage::AnnualUsageBenchmarksService` which uses some anonymised school data. To make the specs less fragile and easier to revise in future, I have revised them to use data created by FactoryBot. This is sufficient for testing that the calculated values are correct.

The PR updates the rubocop config to include the changed files, so can chip away at the rest of the code.


urbanwide (approved):
Looks good @ldodds. Good move with rubocop too 👍

urbanwide (approved):
Looks good @ldodds 👍

PR #635, opened 2023-11-22
Fix calculation of `BenchmarkMetrics.benchmark_energy_usage_£_per_pupil` for storage heaters

ldodds:
The BenchmarkMetrics class calculates various benchmarking figures for schools.

The `benchmark_energy_usage_£_per_pupil` method calculates a total expected per-pupil energy cost for a list of schools, for either well managed or exemplar schools.

Investigating a problem with generating some benchmark variables I found that the alert that uses this method is failing for schools with storage heaters. The bug was introduced in [this change last year](https://github.com/Energy-Sparks/energy-sparks_analytics/commit/65779f7f97ef19adc15849f56ef865d75dbf0597#diff-e4dd4e5bd5af7604cefb6b6617e0eb0eb93e0d56a7eefb3cf4b91de1cd441e53).

When attempting to calculate the expected per-pupil heating cost for storage heaters it was attempting to use the gas costs for the school. But the school won't necessarily have a gas meter. Elsewhere when calculating storage heater costs we use the electricity costs, as storage heaters are electrically powered.

This PR does a couple of things:

- it moves creation of the school activation date variable used in the school comparison pages to another alert, that is less likely to fail because of issues with limited data or calculation errors for a school. It was this missing variable which highlighted the bug
- refactors the BenchmarkMetrics class to make a number of methods private, this helps simplify the public interface
- fixes the bug in calculating the storage heater costs by using the electricity costs instead
- runs rubocop over the affected files

I still think there are some bugs in `AlertEnergyAnnualVersusBenchmark` (it still fails for some schools) and possibly `BenchmarkMetrics` (I'm not sure the variables being used in the gas/storage heater calculations are correct, but these have been in place for several years). Fixing those will happen in follow-up PRs but needs further investigation.

This fix is higher priority as the missing variables will mean our impact reporting is not fully complete.


urbanwide (approved):
Looks good @ldodds 👍

PR #636, opened 2023-11-22
Update climate change levy key and figures

ldodds:
When the costs for the climate change levy are added to the calculated cost breakdown they currently use a different key for each year, e.g. `:climate_change_levy__22_23`. This means every year we need to update the application with new translation keys otherwise the costs are missing from the table.

There's no real need to have different line items for each version of the climate change levy, so this PR changes the code to just use a single key: `:climate_change_levy`.

I've also updated the class to include the costs for 2024/25.

Aside: there's some behaviour in the class which I've left in for now. It looks to see if we have missing configuration for recent years and then throws an exception. But for an older year it just defaults the rate to zero. The former is useful to catch problems where we've not updated the figures, but seemed a bit odd to have an exception and a default.

### DIFF
--- a/app/models/tariffs/climate_change_levy.rb
+++ b/app/models/tariffs/climate_change_levy.rb
@@ -1,31 +1,36 @@
+# frozen_string_literal: true
+
 require 'date'
 
 class ClimateChangeLevy
   class MissingClimateChangeLevyData < StandardError; end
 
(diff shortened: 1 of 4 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #637, opened 2023-11-27
try ruby 3

tbhi:


ldodds:
Marked as blocker so we don't merge for the moment.

Otherwise PR looks good. We can merge this down as part of coordinating the final ruby 3 upgrade.

ldodds (approved):

ldodds:
Am going to merge this as I want to create new tagged version of analytics for upgrading app to Ruby 3.0.

But I was wondering whether we should have the Gemfile.lock reinstated, maybe in favour of the gemspec? We don't actually release this as a gem.

PR #638, opened 2023-11-27
try current faraday

tbhi:
- update to faraday 2
- refactor/rubocop test
- add a test for retry behaviour


ldodds:
@tbhi is this not independent of Ruby 3? If so we could merge this now. It'll get released to test server for at least 24 hours which'll allow for a full run-through of the Meteostat load.

tbhi:
it probably could be but I used the ruby 3 branch as the base so would need some tweaks

PR #640, opened 2023-11-29
Revise OVO benchmark configuration, disable normalisation

ldodds:
The framework for the arbitrary period comparisons applies a set of normalisations to the data for the previous period:

- calculates average weekday and weekend usage figures and then scales those to the number of weekdays/weekends in the current period. This allows the periods being compared to have different lengths, e.g. summer holiday and autumn half term
- applies temperature compensation using the average temperature for the current period
- adjusted those values further depending on changes to the pupil numbers and floor areas in the school

While these are helpful in other contexts, for the Ovo report we just want to see a comparison against current and previous year without these adjustments. For a Jan-August period the temperature compensation may not be correct and the normalisation isn't needed as the period lengths are the same.

This PR adjusts the alerts that generate the variables plus the configuration.

I've tested this locally with the test_benchmarks.rb script to ensure it works as expected.

I've not added additional tests around this as we'll be reworking this code in the coming months.


urbanwide (approved):
Looks good @ldodds 👍

PR #653, opened 2024-01-05
Remove simulator prototype

ldodds:
Further tidying up of the analytics project. This PR removes the prototype electricity usage and building heat simulators which was never fully finished or integrated into the application. These are no longer required.


urbanwide (approved):
Looks like a good tidy up 👍

PR #655, opened 2024-01-10
Qualify series names for meter breakdown charts

ldodds:
There are a variety of "series manager" classes that are responsible for building the time series that are added to charts. The `MeterBreakdown` series is responsible for building series for charts that graph data for individual meters.

Currently there's a bug as if multiple meters of the same fuel type have the same name, then one series overwrites the others. This was introduced when we simplified the series names for meters to use the user supplied name rather than always including the MPAN/MPRNs.

This PR fixes that issues by qualifying the series names to add the MPAN/MPRN if there's a naming clash.

I've also slightly reworked a few naming/labeling methods on the `Dashboard::Meter` class. There as a `meter_no` attribute which was used to conditionally build names, but its not being set anywhere in the code. So probably a hold-over from earlier versions. Have removed it and simplified the methods.

### DIFF
--- a/spec/lib/dashboard/charting_and_reports/charts/series/meter_breakdown_spec.rb
+++ b/spec/lib/dashboard/charting_and_reports/charts/series/meter_breakdown_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Series::MeterBreakdown do
+  let(:meters) { build_list(:meter, 3) }
+  let(:meter_collection) { build(:meter_collection, :with_aggregate_meter, :with_electricity_meters, meters: meters) }
+
+  let(:chart_config) do
+    {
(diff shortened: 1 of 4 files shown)


spec/lib/dashboard/charting_and_reports/charts/series/meter_breakdown_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:meter_collection) { build(:meter_collection, :with_aggregate_meter, :with_electricity_meters, meters: meters) }
+
+  let(:chart_config) do
+    {
+      meter_definition: :allelectricity,
+      series_breakdown: :meter
+    }
+  end
+
+  subject(:series_manager) do
```
tbhi, line 16: subject first? https://rspec.rubystyle.guide/#leading-subject

tbhi (approved):

PR #658, opened 2024-01-12
Fix navigation issues on weekly breakdown charts

ldodds:
Fixes 2 issues on the weekly breakdown charts, e.g. as used on the school dashboards:

- navigating backwards to a previous year breaks if there's less than a week's worth of data. If there's more data than that then the chart x-axis automatically adjusts, but with only a few days we can't display the data. Fix involved first ensuring that the same periods calculations are used across the code that decides whether navigation is possible and the code that calculates the x-axis periods. Then a minimum period size was introduced so user isn't given an option to navigate if we can't display the chart

- when we align the weekly series on the chart we use a Sunday-Saturday period. Currently the code just pushes the data back to a saturday without checking if the end date is already as saturday. That means we lose a week on the charts. Updated code to only adjust end date if needed

### DIFF
--- a/spec/lib/dashboard/utilities/holidays_spec.rb
+++ b/spec/lib/dashboard/utilities/holidays_spec.rb
@@ -9,7 +9,7 @@ describe Holidays do
     let(:minimum_days) { nil }
 
     let(:periods) do
-      Holidays.periods_cadence(
+      described_class.periods_cadence(
         start_date,
         end_date,
         include_partial_period: include_partial_period,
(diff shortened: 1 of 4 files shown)


spec/lib/dashboard/utilities/holidays_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      it 'returns single period' do
+        expect(periods.length).to eq 1
+      end
+
+      context 'when partial periods allowed' do
+        let(:include_partial_period) { true }
+
+        it 'returns two periods' do
+          expect(periods.length).to eq 2
```
tbhi, line 58: worth checking the dates?
```
expect(periods).to contain_exactly(
  have_attributes(start_date: Date.new(2023, 1, 3), end_date: end_date),
  have_attributes(start_date: start_date, end_date: start_date)
)
```

tbhi (approved):

PR #659, opened 2024-01-12
Remove and rearrange old code in charting

ldodds:
This PR:

- Removes some unused classes and code that still existed in the `charting_and_reports` directory
- Moves code from `lib/dashboard/charting_and_reports/test_support` to `test_support`
- Moves code from `lib/dashboard/charting_and_reports/equivalences` to `lib/dashboard/equivalences`
- Moves various files from `lib/dashboard/charting_and_reports` to better location, e.g. `utilities`, `alerts`, etc
- Moves all files from `lib/dashboard/charting_and_reports/old_advice` to `lib/dashboard/advice` with a README to note that this is be retired eventually
- Moves some other code that runs as part of alert framework or referenced from services to be in more appropriate locations
- Renames `ScalarkWhCO2CostValues` to `CalculateAggregateValues` which is a clearer name for that service

This significantly tidies the `lib/dashboard/charting_and_reports` directory. A second round of updates can rename `charting_and_reports` to `charts` and reorganise the remaining code into properly named sub_folders and separate files.


app/services/calculate_monthly_track_and_trace_data.rb
```diff
@@ -70,8 +72,8 @@ def kwhs_for_date_ranges(date_ranges, meter, target_start_date)
       if date_range.nil? || target_start_date > date_range.last
         nil
       else
-        month_start = [target_start_date, date_range.first].max
-        month_end   = [target_start_date, date_range.last ].max
+        [target_start_date, date_range.first].max
```
tbhi, line 75: what's this for?

tbhi (approved):

app/services/calculate_monthly_track_and_trace_data.rb
ldodds: A change made by Rubocop. The variables weren't being assigned and its just removed the assignment but left in the method calls. It looks like those lines should just be removed, will double check there's no side effects.

PR #662, opened 2024-01-16
Rework AnnualUsageBreakdownService

ldodds:
This PR reworks how we calculate the out of hours usage to allow the advice pages to be more flexible when a school has limited data. We want to be able to calculate out of hours usage for shorter time periods, but won't be able to benchmark usage which requires a full year of data.

Currently the calculations in the advice pages are done using the `AnnualUsageBreakdownService`. I've created a new service `UsageBreakdownService` to work with a more flexible date ranges. The `AnnualUsageBreakdownService` has been refactored to delegate to that class, retaining a backwards compatible interface until we update the application.

The code in `AnnualUsageBreakdownService` was created in early 2023 was written by extracting code from `AlertOutOfHoursBaseUsage`. That original code delegated calculation of the usage breakdown across a year to the charting framework by running some pre-configured charts via the `ChartManager`. 

In the new implementation I've completely removed this dependency on the charting framework. It added complexity to the code and obscured the underlying calculations which are otherwise fairly straight-forward (looping through days, calculating a total usage breakdown).

I've also added a new set of specs use FactoryBot. That has required revising some of the existing factories. 

Those specs now check that we can calculate the desired values across all 3 fuel types, across different time periods and with/without community use periods. This improves on what we currently have for `AnnualUsageBreakdownService` (which only tests electricity and storage heaters, for a 1 year periods with no community use). I've left those in place for the moment.

There is also currently unused code in the `AnnualUsageBreakdown` class that is capable of producing estimated savings if the school adjusts their out of hours usage to become a "well managed" or "exemplar" school. This can only be calculated if a school has >1 year of data, so should only be used in that case.

The above steps should allow this code to be merged ahead of switching the application code over to use the new services on the advice pages as all the currently used functionality should be backwards compatible.

A later refactoring will involve changing the `AlertOutOfHoursBaseUsage` which still uses the original code to instead use the new services. This will simplify the codebase further and allow the 12 pre-configured charts to be removed from `ChartManager`.


spec/app/services/usage/usage_breakdown_service_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+
+    subject(:day_type_breakdown) { service.usage_breakdown }
+
+    it 'returns the totals' do
+      expect(day_type_breakdown.total.kwh).to eq expected_total_kwh
+      expect(day_type_breakdown.total.co2).to eq days * 48 * carbon_intensity * usage_per_hh
+    end
+
+    it 'returns the out of hours usage analysis' do
+      weekday_closed_usage_kwh = 11 * (48 - 16) * usage_per_hh
```
tbhi, line 203: seems like these are repeated and could be functions, thinking like
```
def weekday_closed_usage_kwh(usage_per_hh)
  11 * (48 - 16) * usage_per_hh
end

etc

def check_analysis(usage, kwh)
  expect(usage.kwh).to have_attributes(kwh: kwh, co2: kwh * carbon_intensity, £: kwh * flat_rate)
  expect(usage.percent).to be_within(0.001).of(kwh / expected_total_kwh)
end
```
also guessing the numbers are obvious but weren't apparently to me? 48 I guess is the meter readings, but 16 and 4?

app/classes/usage/usage_breakdown.rb
```diff
@@ -0,0 +1,51 @@
...
+      CombinedUsageMetric.new(
+        kwh: total_annual_kwh,
+        co2: total_annual_co2,
+        £: total_annual_£
+      )
+    end
+
+    private
+
+    def total_annual_£
```
tbhi, line 35: worth having something like?
```
    def periods_total(metric)
      [@holiday, @weekend, @school_day_open, @school_day_closed, @community].sum { |period| period.public_send(metric) }
    end
```

tbhi (approved):

spec/app/services/usage/usage_breakdown_service_spec.rb
ldodds: I originally had these as a single set of let statements with comments noting what the numbers were. But had to inline them to stay within the let limit. I'll use a function instead.

PR #668, opened 2024-01-25
allow peak usage calculations and benchmarks to work with limited data

tbhi:


app/services/usage/peak_usage_benchmarking_service.rb
```diff
@@ -31,20 +31,8 @@ def estimated_savings(versus: :exemplar_school)
       end
     end
 
-    def enough_data?
```
ldodds, line 34: In the application you're not referencing the constant in the alert. Wouldn't it make sense to do that here too? There is a minimum threshold of 60 days.

An alternative approach is to retain the methods and the application side changes and update this code so it references the constant in the alert

app/services/usage/peak_usage_calculation_service.rb
```diff
@@ -16,20 +16,13 @@ def average_peak_kw
       peak_kws.sum / peak_kws.length
     end
 
-    def enough_data?
```
ldodds, line 19: Similar comment as above, there's no now requirement for a minimum data, but we're aligning around 60 days?

app/services/usage/peak_usage_benchmarking_service.rb
tbhi: not sure of the benefit of `enough_data?` here rather than `PeakUsageService`?

ldodds (approved):

PR #670, opened 2024-02-05
Add new 12 month period for chart and ensure monthly comparisons align correctly

ldodds:
The `electricity_by_month_year_0_1` and `gas_by_month_year_0_1` charts on the long term usage pages show a monthly comparison over a rolling 2 years.

These charts were recently changed to use a `up_to_a_year` timescale, rather than `year` to make them more flexible. However this has exposed a bug which is resulting in the data for the two periods not displaying correctly. E.g. if there is data for 18 months between June 2022 and December 2023 then the 6 month period for June 2022 should align with June-December 2023.  Instead its showing as Jan-June. 

See example below for a 16 month period. In this example the 4 months from the previous year should be shown as Oct-Jan.

image

Each chart series is actually calculated separately then merged to create the final results. There is code in `AggregatorMultiSchoolsPeriods` which does this. It handles combining series for multiple schools and/or comparisons across multiple years. 

There turned out to be two bugs that code:

- The cause of the above error was that the code was only looking for charts with a `month` x-axis and not `month_excluding_year`. So the series merging wasn't being applied. This was simple change to the logic in `AggregatorConfig` which has a method to determine whether a monthly annual comparison is being carried out.
- I then discovered another bug that caused the merged data to be incorrect if there were duplicate month names (e.g. Jan-Jan as shown in example above). Some 52*7 week periods cover 13 rather than 12 months, so this can occur quite frequently. For example [52 weeks before 2024-01-31 covers 12 months](https://calculat.io/en/date/count-day/52--weeks--before--january--31) (Feb-Jan) but [52 weeks before 2024-02-01 covers 13 months](https://calculat.io/en/date/count-day/52--weeks--before--february--1) (Feb-Feb). Fixing that required reworking the merging code.
 
I've added specs around all those changes.

There's another related bug for the `month_excluding_year` period if it's combined with a `year` or `up_to_a_year` period. The calculation of daily totals in `AggregatorSingleSeries.aggregate_by_day` assumes there's no duplicate columns. So when there are multiple Januaries, all the usage is added to one month, rather than being split across two columns. I've not fixed that and have just I've left a comment about this in the code.

Because there was a requirement to change the `electricity_by_month_year_0_1` and `gas_by_month_year_0_1` charts to compare 2 x 12 months periods, rather than 2 x 52 week periods, I've added a new `twelve_months` time period to do that and reconfigured the charts to use that instead, rather than fixing the other bug. This means that the other bug won't be triggered now as we're no longer using `month_excluding_year`. I may end up just removing support for that.

A new shared context has been added to bring together some of the shared objects useful for testing post-aggregation code in the analytics.


urbanwide (approved):
Looks good on first pass 👍

PR #673, opened 2024-02-20
Provide better method for accessing variables for reports

ldodds:
The alert base class provides a few methods for accessing the variables produced during the analysis.

- `raw_template_data` which returns the variables unmodified
- `front_end_template_data` which returns a formatted version of the variables, ignoring tables, charts and boolean values (not sure why the latter is removed)
- `benchmark_template_data` which returns a subset of the variables, renaming them with short codes
- ...etc

For the new school comparison framework we want to just store all of the unformatted variables, ignoring the chart and table variables. This PR provides a new method, `variables_for_reporting`, that access to these as this combination isn't available otherwise.

The method includes a special case for handling Ranges (mapping them to a low/high value). This is intended to ensure that the unformatted variables we store for comparisons match those returned by `front_end_template_data`.

### DIFF
--- a/lib/dashboard/charting_and_reports/content_base.rb
+++ b/lib/dashboard/charting_and_reports/content_base.rb
@@ -228,16 +228,39 @@ class ContentBase
     list.merge(convert_range_template_data_to_high_low(list, lookup, raw_data))
   end
 
+  # Called by the application to save variables and values used for building reports,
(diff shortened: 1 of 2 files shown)


urbanwide (approved):
Looks good @ldodds 👍

PR #674, opened 2024-02-28
Watt seems clearer? or at least W rather than w?

tbhi:


PhilipTB:
technically watts is not an SI unit W is, as is m2 but not metres squared. It was wrong before because it W is upper case and not lower case.

PhilipTB:
I should have said symbol and not unit. This perhaps needs wider discussion in the group? we use kWh but not kilowatt-hour, I can understand why W may not be as clear as watt to the casual observer

config/locales/benchmarking/content_base.yml
```diff
@@ -301,7 +301,7 @@ en:
           unadjusted_kwh: Unadjusted change (kWh)
           urn: URN
           variation_in_baseload_between_days_of_week: Variation in baseload between days of week
-          w_floor_area: w/floor area
```
ldodds, line 304: I'd suggest W/floor area rather than Watt.

ldodds (approved):

config/locales/benchmarking/content_base.yml
tbhi: W/m²?

PR #675, opened 2024-02-29
changes for change in electricity use between this holiday and the same holiday last year 

tbhi:


tbhi:
will just use html_safe for now

ldodds (approved):

PR #678, opened 2024-03-13
Add html keys for all percentage unit types

urbanwide:
These were found to be missing by doing this: https://github.com/Energy-Sparks/energy-sparks/pull/3497


tbhi:
> Pondering why there would be no Gemfile.lock in the repo?

I removed it a while ago as I'm used to gems not having one - as the application has it.

Looking at https://bundler.io/guides/faq.html "Q: Should I commit my Gemfile.lock when writing a gem?" and "Q: But I have read that gems should not check in the Gemfile.lock!" that seems out of date though.  Perhaps also as this is not just a library.

urbanwide:
Thanks @tbhi - I've not actually checked anything in to this repo before but can see why it's like this now, thanks 👍

ldodds (approved):

PR #681, opened 2024-03-20
Add last_year_holiday_start_date, last_year_holiday_end_date and last_year_holiday_type to AlertImpendingHoliday

urbanwide:


tbhi:
I like the method simplification but probably would have gone for more methods like the `name_of_last_year_holiday` one to avoid extra instance variables

urbanwide:
Thanks @tbhi - was just going with the same format as is there already for consistency 👍

PR #683, opened 2024-03-25
Rework arbitrary period comparison alert to allow periods to be provided as parameters

tbhi:


spec/factories/target_school_factory.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :target_school, class: 'TargetSchool' do
```
ldodds, line 4: The TargetSchool class is used within the targets and tracking feature only. It extends MeterCollection to change how the meters are created (using a TargetMeter which has future projections of fuel consumption), So isn't the normal object that would be used by an alert.

It obviously extends MeterCollection so can be used interchangeably, but I'd expect the specs to create a MeterCollection to avoid hitting any custom behaviour in the targets. We're also likely to remove the TargetSchool/Meter and related classes in an upcoming release.

lib/dashboard/alerts/time period comparison/alert_configurable_period_electricity_comparison.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require_relative 'alert_arbitrary_period_configurable_comparison_mixin'
+
+class AlertConfigurablePeriodElectricityComparison < AlertArbitraryPeriodComparisonElectricityBase
```
ldodds, line 5: approach looks good to me here.

spec/lib/dashboard/alerts/time period comparison/alert_configurable_period_electricity_comparison_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require 'spec_helper'
+
+describe AlertConfigurablePeriodElectricityComparison do
+  let(:alert) do
+    described_class.new(build(:target_school, start_date: Date.new(2022, 11, 1),
+                                              end_date: Date.new(2023, 11, 30)))
+  end
+
+  describe '#calculate' do
```
ldodds, line 11: would be useful to expand this to cover some of the other parameters, e.g. enough days data, max days out of date, etc.

spec/factories/target_school_factory.rb
tbhi: changed to meter collection (confusion as parameter is often named school)

spec/lib/dashboard/alerts/time period comparison/alert_configurable_period_electricity_comparison_spec.rb
tbhi: added

spec/lib/dashboard/alerts/time period comparison/alert_configurable_period_electricity_comparison_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      enough_days_data: 1,
+      current_period: Date.new(2023, 11, 24)..Date.new(2023, 11, 24),
+      previous_period: Date.new(2023, 11, 17)..Date.new(2023, 11, 17)
+    }
+  end
+
+  describe '#analyse' do
+    it 'runs and sets variables' do
+      alert.analyse(Date.new(2023, 11, 30), comparison_configuration: configuration)
+      expect(alert.analysis_date).to eq(Date.new(2023, 11, 30))
```
ldodds, line 25: You could also add expects for `current_period_start_date`, `current_period_end_date`, etc to confirm that the supplied dates have been used.

ldodds (approved):

spec/factories/meter_collection_factory.rb
```diff
@@ -30,6 +30,39 @@
       end
     end
 
+    trait :with_fuel_and_aggregate_meters do
+      fuel_type { :electricity }
+      transient do
+        pseudo_meter_attributes do
```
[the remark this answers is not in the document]
tbhi: removed

ldodds (approved):

PR #685, opened 2024-04-12
Reworking usage during current holiday alert

ldodds:
We have alerts that trigger during holidays to warn schools if they are consuming energy. They calculate the amount consumed so far in the holiday period and used that to give a projected consumption for the whole holiday.

There are several issues with the current gas and storage heater versions of these alerts:

1. at the start of the holiday we might only have 1-2 weekend or week days. This means the projected consumption calculations underestimate the actual consumption as it can't calculate the average weekday or weekend consumption
2. the alerts build a heating model to calculate the number of days when the heating or hot water is on. These are provided as variables but aren't actually used. The need for a heating model means that some schools with limited data, or problems generating a model, won't get this alert
3. the gas alert is also set to be `never_relevant` for a school whose gas usage is configured as being in the kitchen only, or just for hot water usage. But we want to make them aware about all usage during holiday periods, not just when there is some heating 

Also, none of the alerts apply a threshold to help flag when consumption is significant. So even a tiny amount of usage will trigger the alert. This means we can end up showing very tiny cost savings, e.g. a few pence, particularly at the start of a holiday when there may only have been weekend usage.
 
The electricity version of the alerts doesn't have the issue with the calculations. That code calculates the consumption and projections differently and falls back on weekend or weekday usage if there is limited data. So that code is a better basis for refactoring the alerts to use a common set of calculations without the need for a heating model.

So the key changes here are:

- extract a new base class for all three alerts from the electricity alert
- refactor the alerts to use that base class, adding tests for each of them
- adding test coverage around `enough_data`, `relevance`, and `rating` and the calculations to ensure the alerts are working correctly
- add a £10 threshold for consumption
- general tidy up of the existing code using rubocop


lib/dashboard/alerts/common/alert_usage_during_current_holiday_base.rb
```diff
@@ -0,0 +1,242 @@
...
+    @holiday_projected_usage_£ = projected_totals[:£]
+    @holiday_projected_usage_co2 = projected_totals[:co2]
+
+    # Ignore usage less than a threshold. A nil rating means the alert calculation
+    # will be ignored by the application
+    @rating = if @holiday_usage_to_date_£ < USAGE_THRESHOLD_£
+      nil
+    else
+      0.0
+    end
```
tbhi, line 161: perhaps `@holiday_usage_to_date_£ < USAGE_THRESHOLD_£ ? nil : 0.0`?

lib/dashboard/alerts/common/alert_usage_during_current_holiday_base.rb
```diff
@@ -0,0 +1,242 @@
...
+  def holiday_start_date
+    @holiday_period&.start_date
+  end
+
+  def holiday_end_date
+    @holiday_period&.end_date
+  end
+
+  def totals(usage_to_date)
+    usage_to_date.transform_values { |v| v.values.map { |vv| vv[:total] }.compact.sum }
```
tbhi, line 201: filter_map would be better?

lib/dashboard/alerts/common/alert_usage_during_current_holiday_base.rb
```diff
@@ -0,0 +1,242 @@
...
+    },
+    heating_type: {
+      description: 'gas boiler or storage heaters, or nil',
+      units: String,
+      benchmark_code: 'ftyp'
+    }
+  }.freeze
+
+  # Check whether we are in a holiday, otherwise delegate to parent
+  def relevance
```
tbhi, line 78: attr_reader looks to duplicate

lib/dashboard/alerts/storage heaters/alert_storage_heater_heating_on_during_holiday.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  protected
+
+  def needs_electricity_data?
+    false
+  end
+
+  def needs_storage_heater_data?
+     true
```
tbhi, line 48: indentation

tbhi (approved):

urbanwide (approved):
Looks good 👍

lib/dashboard/alerts/common/alert_usage_during_current_holiday_base.rb
ldodds: Wasn't sure what you meant, but have reduced to `usage_to_date.transform_values { |v| v.values.map { |vv| vv[:total] }.sum }`

PR #694, opened 2024-05-13
Fix solar charts when school has storage heaters

ldodds:
When a school is configured to have both solar and storage heaters, the monthly solar chart ends up not including any mains consumption. Without the storage heater configuration the chart displays correctly, but after adding the storage heater there is no mains consumption series.

There turned out to be several issues here.

### Chart filtering

The initial bug was caused by the the code that filters out series from the solar charts. That code expected the meters shown on that chart to have specific hard-coded names. 

However, this breaks when we have both solar and storage because the mains electricity meter gets renamed and no longer matches the expected string.

There was a previous incomplete change to move the aggregation code away from using fixed names and towards having named types (e.g. `:export`, `:mains_consume`) for the sub meters. But this wasn't applied to the chart framework, which continues to look for hard-coded names. 

So the initial changes in this PR rework the chart filtering code such that the series filtering is based on the submeter types and not their names. These changes involve:

- decomposing the old `AggregationFilter` class into two smaller classes `Charts::Filters::DateFilter` and `Charts::Filters::SeriesFilter`. These do separate types of filtering and are called in different ways
- add tests around all of the above, revising the rest of the framework to use the new classes
- revising the `Charts::Filters::SeriesFilter` class to accept a list of sub_meter types rather than meter names. This was the main functional change here
- updating the chart configuration to work with the revised filter, this was the final part of the fix.

This seemed to initially fix the charts. But exposes additional issues. 

### Incorrect mains consumption

The storage heater aggregation code occurs after the solar processing. This ends up replacing some electricity meters with synthetic versions that break out the storage heater usage.

If a school also has solar then the storage heater disaggregation is applies to a previously created synthetic meter that corresponds to the mains + self consumption from the original electricity meter.

But the storage heater processing doesn't correctly check for this, which means that when the above charts are working, what is displayed is a "mains consumption" series which actually consists of the mains consumption + self consumption.

The fix here was to change the `:mains_consume` sub meter relationship to always point to the original mains electricity meter and not the meter created in the solar aggregation.

### Incorrect meter ids

Finally, after testing schools with both solar and storage it turned out that the tariff charts were breaking. This turned out to be due to how the sub meters are combined as a final step that occurs after the solar and storage processing.

The code was breaking the contract that the id and mpan_mprn for a Dashboard::Meter should match. 

This presumably wasn't an issue before, but was exposed by the above changes. 

In addition to adding a comprehensive set of tests around the solar and storage heater aggregation code I've also:

- added some additional comments to the code, to help clarify what is happening
- tidied up some of the logging calls to use `log.debug {...}` to optimise memory usage on production
- manually reviewed chart data and the full set of advice pages for a range of schools (with/without solar, with/without storage) to identify any further issues. There's one issue with storage heater meter identifers which is not related to these changes which I'll address separately.

I've added comments around the specific functional changes.


lib/dashboard/charting_and_reports/charts/filters/series_filter.rb
```diff
@@ -0,0 +1,114 @@
...
+      # This method filters out sub meters based on their type. If there's no filter we
+      # just keep all the series, otherwise we only keep those series whose meter is of a specific
+      # sub_meter type (e.g. mains_consume, export, self_consume, etc)
+      def submeter_filter(to_keep)
+        return to_keep unless @chart_config.series_breakdown == :submeter
+
+        # this is the meter whose sub meters are being displayed
+        parent_meter = @results.series_manager.meter
+
+        to_keep += if @chart_config.submeter_filter?
```
ldodds, line 45: This is the new submeter filter implementation. Rather than accepting a list of meter names, it accepts a list of submeter types. The submeters that match that list, are kept.

The chart series use the meter names.

app/services/aggregate_data_service.rb
```diff
@@ -205,10 +209,12 @@ def aggregate_sub_meters_by_type(combined_meter, meters)
       log "    Combining type #{sub_meter_type} for #{sub_meters.map(&:to_s).join(' ')}"
       AggregateDataServiceSolar.new(@meter_collection).backfill_meters_with_zeros(sub_meters, combined_meter.amr_data.start_date)
       combined_sub_meter = aggregate_meters(nil, sub_meters, sub_meters[0].fuel_type)
-      combined_sub_meter.id   = sub_meters[0].id
```
ldodds, line 208: Removed the call to override the id here. Normally the id and the mpan_mprn attribute for a Dashboard::Meter are the same, this line broke that causing the `Meter.meter?` method for finding meters to fail.

app/services/aggregation_service_storage_heaters.rb
```diff
@@ -88,37 +118,60 @@ def disaggregate_storage_heat_meter(meter)
...
+    # Copy the solar sub meters from the original electricity meters to be sub meters
+    # of the new version that doesn't include storage heater usage
+    map[:ex_storage_heater].sub_meters.merge!(map[:original].sub_meters)
+
+    # By default the mains consumption meter for the new electricity meter should be
+    # the original meter. However if the original meter was created during the solar
+    # aggregation, then its AMR data will correspond to main_consume + self_consume.
+    # In this case we should use its original mains consumption meter and not the
+    # meter created in the solar aggregation step
+    map[:ex_storage_heater].sub_meters[:mains_consume] = if map[:original].sub_meters.key?(:self_consume) && map[:original].sub_meters.key?(:mains_consume)
```
ldodds, line 158: This is the functional change that ensures that we retain references to the original electricity meter to avoid showing incorrect series on solar charts.

app/services/aggregate_data_service.rb
```diff
@@ -184,10 +182,16 @@ def combine_solar_pv_submeters_into_aggregate
     aggregate_sub_meters_by_type(aggregate_meter, @meter_collection.electricity_meters)
   end
 
-  # Creates a new aggregate meter that combines together all of the mains consumption (and other subtypes)
-  # sub meters associated with the list of meters.
+  # Updated the +combined_meter+ so that it has the same types of sub_meters as the meters in
```
tbhi, line 185: update?

spec/lib/dashboard/charting_and_reports/charts/filters/series_filter_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+                mains_consume
+                export
+                self_consume
+              ]
+            }
+          }
+        end
+
+        it 'filters to the submeters' do
+          filter.filter
```
urbanwide, line 80: I'd be tempted to move the call to filter.filter (and others in the same file) to a before block in the context 'when filtering solar submeters'.

urbanwide (approved):
Looks good - just one comment - seen inline 👍

spec/app/services/aggregate_data_service_storage_heaters_spec.rb
```diff
@@ -0,0 +1,248 @@
...
+
+describe AggregateDataServiceStorageHeaters do
+  subject(:service) { described_class.new(meter_collection) }
+
+  let(:meter_collection) do
+    build(:meter_collection)
+  end
+
+  let!(:electricity_meter) do
+    kwh_data_x48 = Array.new(48, 1.0)
```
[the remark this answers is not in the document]
ldodds: Done

PR #695, opened 2024-05-16
Fix MultipleFuels charts when date periods don't align

ldodds:
Screenshot from 2024-05-16 16-21-59

The above chart (`stacked_all_years`) should show a breakdown of all fuels across all available data.

However if the date ranges for the fuel types don't align then the chart can break. 

E.g. if a school had storage heaters in the past but no longer has them. Then the storage heater date only covers that historical period.

PR just introduces a check to see if we have consumption date for the specified date before retrieving the kwh/£/co2 values.


spec/lib/dashboard/charting_and_reports/charts/series/multiple_fuels_spec.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Series::MultipleFuels do
+  subject(:series_manager) do
+    Series::MultipleFuels.new(meter_collection, chart_config)
```
tbhi, line 7: use described_class?

spec/lib/dashboard/charting_and_reports/charts/series/multiple_fuels_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    }
+  end
+
+  before do
+    meter_collection.add_electricity_meter(electricity)
+    meter_collection.add_heat_meter(gas)
+    AggregateDataService.new(meter_collection).aggregate_heat_and_electricity_meters
+  end
+
+  describe '#get_data' do
```
tbhi, line 43: should be #day_breakdown ?

tbhi (approved):

PR #704, opened 2024-06-20
reverse tariff list as looks like we expect ascending order

tbhi:
`EconomicTariffsChangeCaveatsService` uses `tariff_change_dates_in_period` which uses `find_tariffs_between_dates` - this returns the tariffs in reverse order.  It appears as though we assume the last item in the list returned is the last date where it is actually the first.

Test updated to have an extra tariff so the order issue is exposed.


ldodds (approved):
Looks OK, I double checked where else this is used and looks like only the caveats service is expecting a specific order.

PR #707, opened 2024-08-02
Fix bug in calculating solar offset co2

ldodds:
When calculating the values to display on the management dashboard the co2 values include an "offset" when the school has solar panels. We reduce their co2 emissions by the amount of co2 that has been saved by consuming energy from the solar panels.

The offset calculation can return nil which cause some errors in the application. One reason it might return nil is if there's more limited solar generation data than there is for electricity consumption.

This can happen in a very specific circumstance:

- school has >1 electricity meter
- the panels are linked to a meter which has lagging data (so quite out of date)
- there is a meter attribute on that meter which says to ignore that end date

We then end up with a solar_pv production meter that has less data than the aggregate meter.

To try and improve the code I've changed the calculation so that it uses an "up_to_a_year" period, rather than "year". This means it'll tolerate less data, allowing us to calculate a better offset.

I've also fixed a bug which meant that we were calculating the co2 emissions twice for schools with solar (calculation was run, then results discarded).

The management table code needs a complete rework so considering this a temporary improvement. 

I've added a spec to ensure that we're testing the solar offset calculation code as it wasn't covered before.

### DIFF
--- a/lib/dashboard/alerts/common/management_summary_table.rb
+++ b/lib/dashboard/alerts/common/management_summary_table.rb
@@ -276,8 +276,11 @@ class ManagementSummaryTable < ContentBase
   def compare_two_periods(fuel_type, period1, period2, max_days_out_of_date)
     current_period_kwh  = checked_get_aggregate(period1, fuel_type, :kwh)
     previous_period_kwh = checked_get_aggregate(period2, fuel_type, :kwh)
-    current_period_co2  = checked_get_aggregate(period1, fuel_type, :co2)
-    current_period_co2  = electricity_co2_with_solar_offset(period1) if @school.solar_pv_panels? && fuel_type == :electricity
+    current_period_co2 = if @school.solar_pv_panels? && fuel_type == :electricity
+      electricity_co2_with_solar_offset(period1)
+    else
+      checked_get_aggregate(period1, fuel_type, :co2)
+    end
     current_period      = checked_get_aggregate(period1, fuel_type, :£)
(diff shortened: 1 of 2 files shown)


lib/dashboard/alerts/common/management_summary_table.rb
```diff
@@ -276,8 +276,11 @@ def difference_to_exemplar_£(actual_£, fuel_type)
   def compare_two_periods(fuel_type, period1, period2, max_days_out_of_date)
     current_period_kwh  = checked_get_aggregate(period1, fuel_type, :kwh)
     previous_period_kwh = checked_get_aggregate(period2, fuel_type, :kwh)
-    current_period_co2  = checked_get_aggregate(period1, fuel_type, :co2)
-    current_period_co2  = electricity_co2_with_solar_offset(period1) if @school.solar_pv_panels? && fuel_type == :electricity
+    current_period_co2 = if @school.solar_pv_panels? && fuel_type == :electricity
+      electricity_co2_with_solar_offset(period1)
+    else
```
tbhi, line 281: think the else would usually align under the if?

tbhi (approved):

tbhi (approved):

lib/dashboard/alerts/common/management_summary_table.rb
ldodds: Yes. I didn't run rubocop over the file to reduce noise in the PR. Have fixed.

PR #711, opened 2024-08-19
Update benchmarking and equivalence constants

tbhi:
- `bev_co2_per_km` is slightly different to spreadsheet - feels like we should continue using the calculated method for consistency?


ldodds (approved):
:+1: agree with leaving ev calculation for the moment

PR #717, opened 2024-09-12
format unit support for non decimal percentage

tbhi:


ldodds (approved):

ldodds:
Maybe add a spec? there's some existing specs for the class

PR #727, opened 2024-10-10
move data sources code to main project

tbhi:


ldodds (approved):
Looks good, but think there are some scripts in `scripts/external` that could be removed as well. I think everything except `check_holidays.rb` uses one of these APIs. Not sure we need to keep them, as they were mostly dev/debugging tools I think.

PR #729, opened 2024-10-18
Add meter attribute to correct overnight solar readings

tbhi:


lib/dashboard/aggregation/validate_amr_data.rb
```diff
@@ -1256,4 +1258,16 @@ def check_temperature_data_covers_gas_meter_data_range
       raise NotEnoughTemperaturedata, "Temperature data from #{@temperatures.start_date} to #{@temperatures.end_date} doesnt cover period of #{@meter.mpxn} #{@meter.fuel_type} meter data from #{@amr_data.start_date} to #{@amr_data.end_date}"
     end
   end
+
+  def override_night_to_zero(rule)
```
ldodds, line 1262: Extract this into a class in the `Corrections` module? Can then also test in isolation.

lib/dashboard/aggregation/validate_amr_data.rb
```diff
@@ -1256,4 +1258,16 @@ def check_temperature_data_covers_gas_meter_data_range
...
+    rule ||= {}
+    start_date = default_start_date(rule[:start_date])
+    end_date = default_end_date(rule[:end_date], false)
+    (start_date..end_date).each do |date|
```
ldodds, line 1266: This will force the readings to zero for the entire range. I think it would be better to only do the substitution if there is non-zero data during the night. If there was no change then we'd retain the ORIG data otherwise substitute with corrected. The meter report could then be used to highlight specific days where there was an issue in the original data, rather than having all the data substituted. 

This would also avoid potential issues with substituting missing solar data. `similar_solar_pv_day` looks for days that are suitable for substituting missing data in the solar feed. I believe it will only using ORIG data (via `original_matching_substitute_date`) when doing this.

As meter attributes are applied in order, this new correction would be applied first, and then the later steps filling missing values. So some otherwise good solar data might become unavailable for substitution.

Alternatively, the `original_matching_substitute_date` data could allow substitution with `SOLN` data, but not sure about implications of that, if we already know there are potential issues with that day.

lib/dashboard/aggregation/validate_amr_data.rb
```diff
@@ -1256,4 +1258,16 @@ def check_temperature_data_covers_gas_meter_data_range
...
+      data = @amr_data.one_days_data_x48(date)
```
ldodds, line 1267: At this point in the validation the may still be gaps in the unvalidated data used to populate the AMR data. So this might return nil. Have confirmed this causes validation to fail with some existing meters/schools.

spec/lib/dashboard/aggregation/validate_amr_data_spec.rb
```diff
@@ -27,20 +34,31 @@
...
+  end
+
+  context 'with override_night_to_zero' do
+    it 'replace night time readings with a rule' do
+      meter.meter_correction_rules << { override_night_to_zero: nil }
+      validator.validate(debug_analysis: true)
+      expect(arbitrary_night_readings(meter)).to eq(Array.new(8, 0.0))
     end
 
-    it 'replaces missing night time solar readings with 0' do
```
ldodds, line 37: Following from above comments, suggest expanding to cover where there are gaps in the amr data, and days that don't need correcting.

ldodds (changes requested):

lib/dashboard/corrections/override_night_to_zero.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Corrections
+  module OverrideNightToZero
+    def self.apply(rule, meter)
+      rule ||= {}
+      p rule
```
ldodds, line 7: debug?

lib/dashboard/corrections/override_night_to_zero.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Corrections
+  module OverrideNightToZero
+    def self.apply(rule, meter)
```
ldodds, line 5: minor, but I'd prefer to see parameters spelled out, rather than relying on hash.

spec/lib/dashboard/aggregation/validate_amr_data_spec.rb
```diff
@@ -27,20 +34,49 @@
...
+      validator.validate(debug_analysis: true)
+      expect(arbitrary_night_readings(meter)).to eq(Array.new(8, 0.0))
+      expect(meter.amr_data.first[1].type).to eq('SOLN')
     end
 
-    it 'replaces missing night time solar readings with 0' do
+    it 'replace night time readings with a rule with dates' do
+      meter.meter_correction_rules[-1] = { override_night_to_zero: { start_date: meter.amr_data.start_date + 1.day,
+                                                                     end_date: meter.amr_data.start_date + 2.days } }
+      p meter.meter_correction_rules
```
ldodds, line 58: debug?

ldodds (approved):

PR #730, opened 2024-10-25
Don't run AlertHeatingComingOnTooEarly during holidays

tbhi:
looking at AlertUsageDuringCurrentHolidayBase for a possible implementation it overrides the relevance method but that doesn't look to be called.  It also puts a return in the calculate.


ldodds:
Not sure if I'm understanding correctly but AnalysisBase.analyse` the calculate method is only called if `valid_alert?` is true. That calls `ContentBase.valid_content?` which in turn checks `relevance`. So calculate shouldn't get called if the alert isn't relevant

The application calls `.relevance` in `AnalyticsAdapter` which is then stored.

ldodds (approved):
Add a spec?

PR #733, opened 2024-10-29
zero negative solar self consumption

tbhi:


ldodds (approved):
Solves the issue. I did wonder if there's a way to do this check in the existing loop that combines the data to avoid looping over the entire date range a second time.

tbhi:
Briefly considered that, probably simpler overall, worth refactoring?  Also is the different reading type useful?

ldodds:
I think it'd be worth trying a refactoring. The method combining the amr data is called from a few places, so any new behaviour would need to be conditional for just when combining solar amr.

As the creation of the solar meters happens in the aggregation step, the resulting data doesn't get written to the database. The meters and their data are just part of the meter collection. So the reading type flag is not as useful here.

ldodds (approved):

PR #735, opened 2024-11-01
Fix errors and remove debug variables in period comparison alerts

ldodds:
Reviewing the alert table size, I noticed that the alerts with the largest records are those that do period comparisons. The configurable period alerts are the largest (>20kb text in `template_data`), but the other period comparisons are also large (> 2kb text in `template_data`).

Checking the output of these alerts I noticed that:

- there was an error thrown when generating the period names which is leading us to store a complete stack trace as the variable value. Repeated twice for each period and once each for the english and welsh formatted variables (`template_data`, `template_data_cy`) and the raw unformatted variables (`variables), this explains why some of the TOAST entries are very large
- there was a missing translation causing another error to appear in the `timescale` variable
- lengthy text to state the tariff has changed during the compared periods which is not used anywhere
- several debug variables which listed every temperature or daily kWh reading for the current and previous periods. Given that some of the periods span 6-9 months thats a lot of extra text

This PR fixes the bugs and removes a number of debug variables which are no longer needed.

These alerts run for most schools, so that's a lot of unnecessary text we're storing. This should greatly reduce the amount of data stored in the alerts TOAST table.

I've regenerated several schools using this version of the code and confirmed the fixes.

### DIFF
--- a/lib/dashboard/alerts/time period comparison/alert_arbitrary_period_comparison_base.rb
+++ b/lib/dashboard/alerts/time period comparison/alert_arbitrary_period_comparison_base.rb
@@ -7,11 +7,15 @@ class AlertArbitraryPeriodComparisonBase < AlertHolidayComparisonBase
   end
 
   private   def period_type;        'arbitrary period'          end
-  protected def period_name(period); basic_configuration[:name] end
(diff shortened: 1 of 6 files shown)


tbhi (approved):
spec possible to recreate the errors?

PR #738, opened 2024-11-04
shorten log messages

tbhi:


lib/dashboard/alerts/common/alert_analysis_base.rb
```diff
@@ -67,8 +67,8 @@ def log_stack_trace(e, backtrace = true)
     @error_message = e.message
     @backtrace = e.backtrace
 
-    logger.warn e.message
-    logger.warn e.backtrace if backtrace
+    logger.warn e.message.truncate(120)
```
ldodds, line 70: could add an ellipsis to indicate its truncated?

ldodds (approved):

lib/dashboard/alerts/common/alert_analysis_base.rb
tbhi: was doing this by default

PR #739, opened 2024-11-04
Fix schoolweek comparison alert

ldodds:
Changes to base class broke the school week gas comparison.

PR fixes the comparison and adds a basic spec for all three variants confirming that they at least run and analyse the right time periods.


ldodds:
There's a certain amount of repetition and we could provide some higher-level factories/helpers for common patterns.

PR #740, opened 2024-11-05
 remove old benchmark code 

tbhi:


ldodds (approved):
Suggest bumping up a major version number when tagging, to make it easier to look back if we need it? (I know we can search the repo too)

PR #745, opened 2024-11-21
don't capture exception message as field value

tbhi:


PhilipTB (commented):
Is this sensible? Won't lots of other reports which need the floor area break without a clear error message, potentially in unexpected ways. Would we 
do better to set a school up with floor area approximately from a satellite photo than returning null?

PR #746, opened 2024-11-22
Update SEG rate used for  solar calculations

tbhi:


ldodds (approved):

ldodds (commented):
This is the fallback value. The actual price is defined by `BenchmarkMetrics.pricing` with default values coming from the application database.

tbhi:
does it need a fallback? perhaps best to remove to simplify?

tbhi:
also surprised this change is changing the capital cost?

ldodds:
These constants were originally in the analytics and then changed to be in the application. The fallback was retained as a safety net, but is there now largely for testing. We'd still need a default value somewhere though if we want to be able to test the alerts/calculations in isolation. So the constant could be replaced with something that populated `BenchmarkMetrics.pricing`

The estimation process looks at the school's annual electricity consumption and then looks at different sizes of solar panels to find best sizing that will offset their consumption costs. So the current export cost does have some bearing on the capital costs and time to recoup that.

While we could change how this constant works for testing purposes, and perhaps enforce prices in the application database, there's no real need for this PR at the moment.

PR #758, opened 2025-01-30
error on formating NaN day

tbhi:
from `Exception in variable_list for Highfields Primary School for AlertHeatingComingOnTooEarly - FloatDomainError: NaN (/var/app/bundle/ruby/3.2.0/bundler/gems/energy-sparks_analytics-f9a9849f112b/lib/dashboard/utilities/format_energy_unit.rb:198:in `to_i')`

### DIFF
--- a/lib/dashboard/modelling/heating/heating_regression_models.rb
+++ b/lib/dashboard/modelling/heating/heating_regression_models.rb
@@ -1055,7 +1055,7 @@ module AnalyseHeatingAndHotWater
     end
 
     def number_of_heating_school_days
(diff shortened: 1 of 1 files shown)


lib/dashboard/modelling/heating/heating_regression_models.rb
```diff
@@ -1055,7 +1055,7 @@ def average_base_temperature
     end
 
     def number_of_heating_school_days
```
ldodds, line 1057: The method is called in other places, e.g. `number_of_heating_days_method_b` in same class where using nil will throw an exception now whereas before the other methods would return `Float::NAN`. So will need to change the other methods as well?

PR #759, opened 2025-02-24
Fallback to not calculating DUoS costs if cannot find region

ldodds:
Non-domestic tariffs include additional time of use charges that vary by time of day and by Distribution Network Operator.

Our current code assumes there are only the existing geographic operators (10-23) but there are now a number of IDNOs which operate across the UK. See [list](https://www.energynetworks.org/customers/find-my-network-operator).

It looks like IDNOs may end up charging the same costs as the local DNO but this isn't 100% clear yet. We're also relying on a set of hard-codes times of day which may not longer be 100% accurate.

As we now have meters that are using an IDNO prefix we need a better fallback than just throwing an exception if we can't find a configuration. Better to omit charges on the tariffs page rather than fail to regenerate the whole school.

So this PR provides a simplistic fallback for meters associated with IDNOS. A better longer term solution will be to:

- find another way to map schools to (I)DNO, e.g. based on postcode prefix
- provide a better way to configure times of use for DUOS charges

### DIFF
--- a/spec/app/models/tariffs/generic_accounting_tariff_spec.rb
+++ b/spec/app/models/tariffs/generic_accounting_tariff_spec.rb
@@ -236,10 +236,6 @@ describe GenericAccountingTariff, :aggregate_failures do
         context 'when the DuoS region is unknown' do
           let(:mpxn)  { 2_712_345_678_900 }
 
-          before do
-            allow(meter).to receive(:mpxn).and_return(2_712_345_678_900)
-          end
-
           it 'includes the zero charges as a fallback' do
             # confirm that we've omitted these costs
             expect(accounting_cost.all_costs_x48[:duos_red]).to be_nil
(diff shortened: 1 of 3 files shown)


spec/app/models/tariffs/generic_accounting_tariff_spec.rb
```diff
@@ -231,6 +232,21 @@
           expect(accounting_cost.all_costs_x48[:duos_amber].sum).not_to be 0.0
           expect(accounting_cost.all_costs_x48[:duos_green].sum).not_to be 0.0
         end
+
+        context 'when the DuoS region is unknown' do
+          let(:mpxn)  { 2_712_345_678_900 }
+
+          before do
+            allow(meter).to receive(:mpxn).and_return(2_712_345_678_900)
```
tbhi, line 240: looks like this is repeated above

tbhi (approved):

PR #761, opened 2025-03-03
Allow navigation on fixed academic year charts

tbhi:


ldodds (approved):
Looks good. Chart subtitle needs fixing in the application as the end date isn't updating as I navigate. Think there's a missing span in the translation.

PR #768, opened 2025-04-04
correct merging on comparison charts

tbhi:
looking through might still be some issues - is this ok?

image


ldodds:
Haven't reviewed the code changes yet. The chart doesn't read right because of the missing months.

But if the issue is that we've only got partial data for the previous academic year then it's not actually worth showing this chart to a user, as they're not going to see any useful monthly comparison. Even with the missing months added in with zero values, the chart won't show anything useful by way of a comparison until the two series overlap from April 2026.

So maybe don't try and show the chart in the application, if there isn't a full previous academic year to display?

ldodds (approved):

PR #770, opened 2025-04-07
don't show comparison  if no months to compare

tbhi:


tbhi:
switched to using a different chart when no comparison

